PR #29, opened 2023-05-22
Add BETAFPVF405_ELRS for FC with onboard UART ELRS

howels:
To better support boards with onboard eLRS and avoid confusion as to which UART to use.


haslinghuis:
Why don't we add this to the F405 target itself?

blckmn (approved):

PR #47, opened 2023-06-03
Update TAKERG4AIO

haslinghuis:
- Rename GEPRCG4AIO to TAKERG4AIO
- Add DPS310
- Add 45 degrees board orientation with USB on top


nerdCopter (approved):
approving, but i cannot find the G4 on GepRC.com

PR #49, opened 2023-06-05
Modify the default receiver protocol of BETAFPVF405_ELRS to CRSF

liyouluo:
This target is for use with the onboard ELRS serial receiver, it is reasonable to use CRSF by default.


haslinghuis:
@liyouluo we have removed the default RX protocol for all targets as this is defined in the cloud build already.

The PR is based off from master instead of a feature branch.

liyouluo:
> @liyouluo we have removed the default RX protocol for all targets as this is defined in the cloud build already.
> 
> The PR is based off from master instead of a feature branch.

I see. I will close this PR.

PR #68, opened 2023-06-13
Add LUXMICROAIO

freasy:


configs/LUXMICROAIO/config.h
```diff
@@ -0,0 +1,123 @@
...
+#define USE_FLASH_W25Q128FV
+#define USE_GYRO_SPI_ICM42688P
+#define USE_SPI_GYRO
+
+#define USE_BARO_DPS310
+#define USE_BARO_BPM280
+#define USE_BARO_BPM085
+#define USE_BARO_MS5611
+#define USE_BARO_QMP6988
+#define USE_BARO_LPS
```
haslinghuis, line 45: Please only specify physical present barometer drivers. Two drivers should be the max allowed.

nerdCopter: @freasy

ot0tot: You can also only define USE_BARO which includes all of the baro drivers by default.

haslinghuis: No this hack is not a recommended practice

mr-sour:
I have one of these boards as they're for sale now. they have DPS310 could we go forward with that so these can be updated.
image

haslinghuis (approved):

configs/LUXMICROAIO/config.h
haslinghuis: Updated the config

mr-sour:
image
<3

PR #88, opened 2023-07-05
Update TRANSTECF7HD

haslinghuis:


nerdCopter (approved):
defining dshot300?

haslinghuis:
Yes manufacturers set this in unified target, so we just follow up. Users still will be able to use ds600 if they like.

PR #101, opened 2023-07-18
Renamed FLYWOOF745 to FLYWOOF745V2

howels:
Name change requested by Flywoo Support.


haslinghuis:
What happens with V1 users - this does not make sense.

howels:
> What happens with V1 users - this does not make sense.

V1 appears to use the FLYWOOF745 target - I will confirm with the vendor

haslinghuis:
V2 should at least change some thing 😛

PR #102, opened 2023-07-20
Fix target for NBD_INFINITYAIO

fabriciopolonio:
Fix target for NBD_INFINITYAIO which currently cannot be built due to missing GYRO and ACC defines.

Issue: https://github.com/betaflight/betaflight/issues/12974


haslinghuis:
Missing define for OSD and FLASH and maybe BARO

fabriciopolonio:
Added USE_OSD, this FC has no barometer.

configs/NBD_INFINITYAIO/config.h
```diff
@@ -69,6 +74,7 @@
 #define DEFAULT_CURRENT_METER_SCALE 230
 #define BEEPER_INVERTED
 #define FLASH_SPI_INSTANCE SPI1
+#define USE_OSD
```
haslinghuis: Need to remove. There are no OSD pins defined. Also this is build option
```suggestion

```

fabriciopolonio: @haslinghuis when you meant OSD you meant the OSD chip?

fabriciopolonio: This FC has no FLASH, BARO or OSD chips.

haslinghuis (approved):

PR #155, opened 2023-08-30
Add a new target config TAKERF411/config

hakrc1:
Add a new target config TAKERF411/config

### DIFF
--- a/configs/TAKERF411/config.h
+++ b/configs/TAKERF411/config.h
@@ -24,7 +24,7 @@
 #define FC_TARGET_MCU                   STM32F411
 
 #define BOARD_NAME                      TAKERF411
-#define MANUFACTURER_ID                 GEPRC
+#define MANUFACTURER_ID                 GEPR
 
 #define USE_ACC
 #define USE_ACC_SPI_ICM42688P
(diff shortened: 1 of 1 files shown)


configs/TAKERF411/config.h
```diff
@@ -0,0 +1,104 @@
...
+ *
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#define FC_TARGET_MCU                   STM32F411
+
+#define BOARD_NAME                      TAKERF411
+#define MANUFACTURER_ID                 GEPRC
```
haslinghuis, line 27:
```suggestion
#define MANUFACTURER_ID                 GEPR
```

hakrc1: The TAKERF411/config.h   is  corresponds to unified-targets\configs\default\GEPRC-TAKERF411.config。if change config.h's MANUFACTURER_ID (from GEPRC to GEPR). Will it cause other problems?

haslinghuis: GEPR is the manufacturer ID.

https://github.com/betaflight/unified-targets/blob/master/Manufacturers.md

haslinghuis (approved):

ot0tot:
Blackbox flash and OSD should not be on the same SPI bus, as it will not be possible to use both at the same time. See: https://github.com/betaflight/betaflight/issues/12583

PR #164, opened 2023-09-06
Cleanup JHEF405PRO target

ot0tot:
* Remove RX_PPM_PIN define because USE_PWM is not defined.
* Default to DSHOT_BURST to OFF because it is incompatible with bidirectional DShot and there are no DMA conflicts that prevent timer-based DShot from working. 
* Remove dashboard I2C define as it is not used.


ot0tot:
May fix https://github.com/betaflight/betaflight/issues/13068#issuecomment-1707639852

haslinghuis (approved):

nerdCopter (approved):

PR #198, opened 2023-10-26
GEPRCF722_BT_HD: Fix the sensor_align to 0

Linjieqiang:
When this board was developed, the BMI270 gyroscope was used. The current mass production gyroscope is ICM-42688, so the inconsistency in the direction of the machine head caused by updating the firmware to the user. The purpose of the modification is to allow the user to maintain the default direction of the firmware after updating the firmware. The FC’s machine heads are in the same direction, reducing user confusion.


nerdCopter:
for config.h , CW0 is default and does not need to be defined. however, it does not hurt either.
https://github.com/betaflight/config/pull/34
https://github.com/betaflight/betaflight/blob/34e7a5ceebb69daa7d730d0c41908385992f3301/src/main/pg/gyrodev.c#L62-L68

nerdCopter (approved):

configs/GEPRCF722_BT_HD/config.h
```diff
@@ -125,6 +125,6 @@
 #define PINIO4_BOX 42
 #define USE_SPI_GYRO
 #define GYRO_1_SPI_INSTANCE SPI1
-#define GYRO_1_ALIGN CW270_DEG
-#define GYRO_1_ALIGN_YAW 2700
+#define GYRO_1_ALIGN CW0_DEG
+#define GYRO_1_ALIGN_YAW 0
```
haslinghuis, line 129: This is default

```suggestion
```

haslinghuis (approved):

PR #201, opened 2023-10-29
Remove USE_BARO defines where no specific hardware is defined

ot0tot:
Per https://discord.com/channels/868013470023548938/924169278926770207/1168257438416375919

One liner: `sed -i '/BARO/d' $(grep -L USE_BARO_ $(grep - lR USE_BARO ./*))`


haslinghuis (approved):

haslinghuis:
Alternative we could use #ifdef USE_BARO block as [GEPRC-M8Q](https://geprc.com/product/gep-m8q-gps-module/) besides GPS would also provide external BARO and MAG

PR #204, opened 2023-10-30
Remove unneeded MAG_I2C_INSTANCE defines

ot0tot:
Remove MAG_I2C_INSTANCE defines from configs where USE_MAG is not defined.


haslinghuis (approved):

McGiverGim:
I'm not too sure how this works, but this does not help if the user decides to use a MAG? If I decide to use a MAG, and I add the define to the custom options of the configurator, if this MAG_I2C_INSTANCE is defined, I suppose I don't need to configure it in the FC. If I don't have this define, I will need to go to the CLI and configure it afterwards. Is that right?

ot0tot:
> I'm not too sure how this works, but this does not help if the user decides to use a MAG? If I decide to use a MAG, and I add the define to the custom options of the configurator, if this MAG_I2C_INSTANCE is defined, I suppose I don't need to configure it in the FC. If I don't have this define, I will need to go to the CLI and configure it afterwards. Is that right?

If you are going to add a MAG, you need to configure it. Defining MAG_I2C_INSTANCE in the config is not desirable, because you don't know which I2C instance the MAG will be connected to. 

I believe the defaults (without MAG_I2C_INSTANCE defined) are:
```set mag_bustype = I2C
set mag_i2c_device = 1
set mag_i2c_address = 0
set mag_spi_device = 0
set mag_hardware = NONE
```

Which should be fine for most users.

haslinghuis:
Alternative we could use `#ifdef USE_MAG` block,

haslinghuis:
@ot0tot please rebase

ot0tot:
> @ot0tot please rebase

I think I resolved all conflicts properly. Please confirm.

PR #234, opened 2023-11-21
Add JHEF411HD

haslinghuis:
Found missing in https://build.betaflight.com/api/builds/474aa96615de7f274b28880e74635e06/log


nerdCopter (approved):

ot0tot:
What a disaster of a FC. If they put OSD and blackbox on SPI3 it might be ok, but SPI2 conflicts with the motor DMA channels. And bitbang is a bad idea due to losing DMA on the gyro. Was this grandfathered in somehow?

PR #235, opened 2023-11-23
AXIS-ARGUSF7AIO:add gyro ICM42688P 

asizon:


configs/ARGUSF7AIO/config.h
```diff
@@ -28,7 +28,10 @@
 
 #define USE_ACC
 #define USE_GYRO
+#define USE_SPI_GYRO
```
haslinghuis, line 31: `USE_SPI_GYRO` i s already defined in target.h

```suggestion
```

haslinghuis (approved):

PR #237, opened 2023-11-24
Revert "Remove unneeded MAG_I2C_INSTANCE defines"

haslinghuis:
Reverts betaflight/config#204

Partially resolves https://github.com/betaflight/betaflight/issues/13181

Reasons:
- needs extra configuration from user when adding mag
- using diff from earlier build does not enable mag because default setting was missing (probably only when I2C port 2 or 3 is used).


nerdCopter (approved):
* i was weary of #204

ot0tot:
What assumptions from here are not correct?
https://github.com/betaflight/config/pull/204#issuecomment-1785391696

haslinghuis:
set mag_i2c_device = 1 was not default in my case.
Boards using port 2 (or 3) have to do an extra step.

haslinghuis:
`set mag_i2c_device = 0` is default.
Boards using port 2 (or 3) have to do an extra step.

PR #239, opened 2023-11-26
Activate dshot telemetry define

haslinghuis:
Uses https://github.com/betaflight/betaflight/pull/13184


haslinghuis:
It uses the same default as unified targets.

PR #241, opened 2023-11-26
True pit for GREATMOUNTAINF435

ASDosjani:
Manufacturer asked for this change. 9V rail is controlled from PD2


configs/GREATMOUNTAINF435/config.h
```diff
@@ -108,6 +108,10 @@
 #define MAX7456_SPI_INSTANCE            SPI2
 #define MAX7456_SPI_CS_PIN              PD5
 
+#define PINIO1_PIN                      PD2
+#define PINIO1_CONFIG 129
```
haslinghuis, line 112:
```suggestion
#define PINIO1_CONFIG                   129
```

PR #252, opened 2023-12-04
Add BMP280 to HAKRCF722D

haslinghuis:


nerdCopter (approved):
removal of ADC also?

haslinghuis:
ADC defines already defined in target.h / adc.h

PR #297, opened 2024-01-13
Readd DARWINF435 target

DMAXYANG:
This should be caused by incorrect LF file, I have revised it again.  
Accidentally shut down the previous PR, so I can only submit again


nerdCopter:
i the future, you may commit to the existing branch and push. the pull request will update accordingly, no need to close and open a new.

nerdCopter (approved):
approving, quick skim, not inspected deeply.

haslinghuis (approved):
USE_ADC is already defined in target.h

nerdCopter (approved):

PR #317, opened 2024-02-06
Mamba H743 2022B - Update ICM gyro version

jesuslg123:
The DIATONE Mamba H743 2022B flight controller it use ICM42688P instead of ICM42605, which seems to have been available on version 1 of the FC.

I am not aware of 2022B (v2) version been sold with ICM42605.


haslinghuis:
Confirmed: DIATONE Mamba H743 2022B only uses MPU6000 and ICM42688P - so we can remove BMI270 also.

haslinghuis (approved):

PR #335, opened 2024-02-22
Add JHEF722HD board config

jhemcu:
Link: https://github.com/betaflight/unified-targets/pull/1258

### DIFF
--- a/configs/JHEF722HD/config.h
+++ b/configs/JHEF722HD/config.h
@@ -1,0 +1,121 @@
+/*
+ * This file is part of Betaflight.
+ *
+ * Betaflight is free software. You can redistribute this software
+ * and/or modify this software under the terms of the GNU General
+ * Public License as published by the Free Software Foundation,
+ * either version 3 of the License, or (at your option) any later
+ * version.
+ *
+ * Betaflight is distributed in the hope that it will be useful,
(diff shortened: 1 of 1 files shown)


configs/JHEF722HD/config.h
```diff
@@ -0,0 +1,121 @@
...
+    TIMER_PIN_MAP( 8, PB3,  1,  0) \
+
+#define ADC3_DMA_OPT 0
+
+#define MAG_I2C_INSTANCE  (I2CDEV_1)
+#define BARO_I2C_INSTANCE (I2CDEV_1)
+
+#define FLASH_SPI_INSTANCE   SPI3
+#define GYRO_1_SPI_INSTANCE  SPI1
+#define GYRO_2_SPI_INSTANCE  SPI1
```
haslinghuis, line 104: SPI sharing is not recommended and can lead to dead locks.

haslinghuis (changes requested):

configs/JHEF722HD/config.h
jhemcu: Thank you for your suggestion. We will avoid this problem in the next board design, but now our products have been launched to the market, and there are no major problems in current use, so I think we can apply for merger.

haslinghuis (approved):

PR #355, opened 2024-03-12
SPRacingH7EVO - Initial target support.

hydra:
For reviewers, this target shares much in common with the H7RF with regards to the schematic, but the form factor is vastly different.

Additionally this FC includes an 8MBps CAN transceiver which will allow the FDCAN peripheral on the H7 to operate at full speed.  `#defines` for which are included here, for when BF adds support for CAN based peripherals which are becoming increasingly common.

```
// 8Mbps CAN transceiver installed with CAN termination, CANL/CANH available on GPS, mezzanine and servo connectors.
#define FDCAN1_RX            PD0
#define FDCAN2_RX            PD1
```


configs/SPRACINGH7EVO/config.h
```diff
@@ -0,0 +1,285 @@
...
+    TIMER_PIN_MAP( 6, PA7 , 2,  3 ) \
+    TIMER_PIN_MAP( 7, PA6 , 1,  2 ) \
+    TIMER_PIN_MAP( 8, PC6 , 2,  8 )
+
+#define DEFAULT_BLACKBOX_DEVICE     BLACKBOX_DEVICE_SDCARD
+
+#define DEFAULT_CURRENT_METER_SOURCE CURRENT_METER_ADC
+#define DEFAULT_VOLTAGE_METER_SOURCE VOLTAGE_METER_ADC
+
+#define USE_SPI_GYRO
```
haslinghuis, line 278: Already defined in target.h

hydra: I'm not quite sure what file you're referring to, I don't have a `target.h` file here.  All the SPRacingH7 FC's have `#define USE_SPI_GYRO` defined, but I do see the following block in `common_post.h`:

```
#if defined(USE_GYRO_SPI_ICM20689) || defined(USE_GYRO_SPI_MPU6000) || defined(USE_GYRO_SPI_MPU6500) || defined(USE_GYRO_SPI_MPU9250) || defined(USE_GYRO_L3GD20) || defined(USE_GYRO_SPI_ICM42605) || defined(USE_GYRO_SPI_ICM42688P) || defined(USE_ACCGYRO_BMI160) || defined(USE_ACCGYRO_BMI270)
#ifndef USE_SPI_GYRO
#define USE_SPI_GYRO
#endif
#endif
```

Are you suggesting that `USE_SPI_GYRO` is no-longer required in all target's `config.h` files?

If so I can remove it from this PR, or, preferrably, leave it in this file (to be consistent), and then remove it from /all/ the remaining targets that still have it as a follow-up PR.

See below:

image

hydra: I'll add a second commit to this PR then.

hydra: done.

configs/SPRACINGH7EVO/config.h
```diff
@@ -0,0 +1,284 @@
...
+#define BUTTON_A_PIN            PC14
+#define BUTTON_A_PIN_INVERTED
+#define BUTTON_B_PIN            PC14
+#define BUTTON_B_PIN_INVERTED
+
+// Device has 2x flash chips in parallel using a single CS pin, initial support is for the first flash chip only.
+#define USE_OCTOSPI
+#define USE_OCTOSPI_DEVICE_1
+
+#define USE_SPI
```
haslinghuis, line 47: This will define `USE_SPI_DEVICE_<x>`

configs/SPRACINGH7EVO/config.h
```diff
@@ -0,0 +1,284 @@
...
+#define I2C1_SCL_PIN            NONE
+#define I2C1_SDA_PIN            NONE
+
+// on GPS connector and mezzanine expansion connector
+// 10k pull-ups installed
+#define USE_I2C_DEVICE_2
+#define I2C2_SCL_PIN            PB10
+#define I2C2_SDA_PIN            PB11
+
+#define I2C3_SCL_PIN            NONE
```
haslinghuis, line 106: NONE is default #ifndef

haslinghuis (changes requested):
Found more :smiley:

configs/SPRACINGH7EVO/config.h
```diff
@@ -0,0 +1,284 @@
...
+#define TARGET_IO_PORTA         0xffff
+#define TARGET_IO_PORTB         (0xffff & ~(BIT(2)|BIT(6)))
+#define TARGET_IO_PORTC         0xffff
+#define TARGET_IO_PORTD         (0xffff & ~(BIT(11)|BIT(12)|BIT(13)))
+#define TARGET_IO_PORTE         (0xffff & ~(BIT(2)|BIT(7)|BIT(8)|BIT(9)|BIT(10)))
+#define TARGET_IO_PORTF         0xffff
+#define TARGET_IO_PORTG         0xffff
+#define TARGET_IO_PORTH         0xffff
+
+#define USE_I2C
```
haslinghuis, line 95: This will define `USE_I2C_DEVICE_<x>`

configs/SPRACINGH7EVO/config.h
hydra: yes, the target definition is being explicit here.  This is how I prefer the SPRacingH7 targets, they're all the same, same goes for the UARTs.

configs/SPRACINGH7EVO/config.h
hydra: Same as situation as `USE_SPI` in my other comment, the target only wants the i2c devices that are defined, to save code space.

See also `STM32H750\target.h`:

```
#if !defined(USE_I2C)
#define USE_I2C
#define USE_I2C_DEVICE_1
#define USE_I2C_DEVICE_2
#define USE_I2C_DEVICE_3
#define USE_I2C_DEVICE_4
#define I2C_FULL_RECONFIGURABILITY
#endif
```

configs/SPRACINGH7EVO/config.h
```diff
@@ -0,0 +1,284 @@
...
+
+#define I2C3_SCL_PIN            NONE
+#define I2C3_SDA_PIN            NONE
+
+#define I2C4_SCL_PIN            NONE
+#define I2C4_SDA_PIN            NONE
+
+
+#define MAG_I2C_INSTANCE        (I2CDEV_2)
+#define USE_BARO
```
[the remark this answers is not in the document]
hydra: fixed

haslinghuis (approved):

nerdCopter (approved):

PR #362, opened 2024-03-24
OMNIBUSF4SD: Add support DPS310 sensor

Linjieqiang:
Alleviating the out-of-stock problem of BMP280 sensors.


haslinghuis (changes requested):
You have to create a new target for your customer. We don't add devices to OEM boards for clones. We did a lot of work to get hardware defines in place so this is counterproductive.

PR #385, opened 2024-04-15
Added configuration for Diatone Mamba MK5 435AI (AT32 Based)

shaneshort:
This is my first pull request, so apologies in advance for any issues! (The doco had a dead link or two so I hope I've followed the correct and current instructions)

This is a target for the Diatone Mamba F435AI flight controller. I requested the defines file for the controller from Diatone (from the AT32 fork of Betaflight) and converted it to Betaflight 4.5 syntax.

I've been flying this exact configuration with RC3 for a few weeks with absolutely 0 issues.

### DIFF
--- a/configs/MAMBAF435_2023A/config.h
+++ b/configs/MAMBAF435_2023A/config.h
@@ -1,0 +1,130 @@
+/*
+ * This file is part of Betaflight.
+ *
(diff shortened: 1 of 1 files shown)


ot0tot:
Does the current sensor work?

PR #398, opened 2024-04-24
Add baro and mag support to GEPRCF722_BT_HD

pinhodb:
GEPRCF722_BT_HD exposes I2CDEV_2 (shared with UART3), commonly used with baro/mag. Add support for DPS310/QMC5883 found in common GNSS modules (e.g. GEP-M10-DQ). 

Verified using Sensors tab in Configurator.


configs/GEPRCF722_BT_HD/config.h
```diff
@@ -35,6 +35,10 @@
 #define USE_GYRO_SPI_ICM42688P
 #define USE_SDCARD
 #define USE_MAX7456
+#define USE_BARO
+#define USE_BARO_DPS310
+#define USE_MAG
+#define USE_MAG_QMC5883
```
haslinghuis, line 41: These are user build options to supply when adding external components to the board when flashing.
Otherwise we would end up adding drivers for all possible configurations not fitting 512K targets.

```suggestion
```

pinhodb: My understanding is these baro/mag devices are very commonly used (with GEPRCF722_BT_HD and others) and board exposes these pins with that in mind.
Here are before/after sizes:

> Memory region         Used Size  Region Size  %age Used
        ITCM_RAM:       15952 B        16 KB     97.36%
      ITCM_FLASH:          0 GB        16 KB      0.00%
ITCM_FLASH_CONFIG:          0 GB        16 KB      0.00%
     ITCM_FLASH1:          0 GB       480 KB      0.00%
      AXIM_FLASH:         480 B        16 KB      2.93%
AXIM_FLASH_CONFIG:          0 GB        16 KB      0.00%
     AXIM_FLASH1:      393815 B       480 KB     80.12%
        DTCM_RAM:       34980 B        64 KB     53.38%
           SRAM1:       57696 B       176 KB     32.01%
           SRAM2:          0 GB        16 KB      0.00%
       MEMORY_B1:          0 GB         0 GB
   text    data     bss     dec     hex filename
 387899    6396   86276  480571   7553b ./obj/main/betaflight_STM32F7X2_GEPRCF722_BT_HD.elf


> Memory region         Used Size  Region Size  %age Used
        ITCM_RAM:       15968 B        16 KB     97.46%
      ITCM_FLASH:          0 GB        16 KB      0.00%
ITCM_FLASH_CONFIG:          0 GB        16 KB      0.00%
     ITCM_FLASH1:          0 GB       480 KB      0.00%
      AXIM_FLASH:         480 B        16 KB      2.93%
AXIM_FLASH_CONFIG:          0 GB        16 KB      0.00%
     AXIM_FLASH1:      380427 B       480 KB     77.40%
        DTCM_RAM:       34968 B        64 KB     53.36%
           SRAM1:       56392 B       176 KB     31.29%
           SRAM2:          0 GB        16 KB      0.00%
       MEMORY_B1:          0 GB         0 GB
   text    data     bss     dec     hex filename
 374591    6316   85040  465947   71c1b ./obj/main/betaflight_STM32F7X2_GEPRCF722_BT_HD.elf

sugaarK:
we can only have the defines for included hardware on board the flight controller. if your smart enough to add an external Baro you can use the define tab

pinhodb:
OK that makes sense. I noticed there are a number of FCs with multiple BAROs in config, but I can see now this is probably because different device can come fitted from manufacturer (compatible footprints).

haslinghuis (approved):

PR #414, opened 2024-05-03
Update SPEEDYBEEF405MINI (remove ESC telemetry)

ot0tot:
Remove ESC telemetry define because this stack ships with a BLHeli_S ESC, and Speedybee does not even make a 20x20 BLHeli32 ESC. 

https://www.speedybee.com/speedybee-f405-mini-bls-35a-20x20-stack/


haslinghuis:
image

ot0tot:
N = Not connected

haslinghuis:
Makes sense to remove from stack configuration - but the FC exposes a specific UART if user would use another ESC

image

ot0tot:
Understood, but lots of FCs have provisions for ESC telemetry and they are not included in the configs.

haslinghuis:
Agree if unused we should not configure it. As long we can find the right port to use in documentation.

haslinghuis (approved):

nerdCopter:
So essentially this just disable the ports-tab pre-configured state, correct?

PR #427, opened 2024-05-17
Add JHEF745V2 target board

jhemcu:
The old version uses the MPU6000 sensor, this version uses the ICM42688 sensor, and the other pin configurations are the same.

b316b2426d764ee561885c2c4d00d69


jhemcu:
You are right. The only difference is the gyroscope...but I think the current compilation system cannot support the #ifdef operation.

haslinghuis (approved):

nerdCopter:
thank you @haslinghuis 

@jhemcu , please note the board name is `JHEF745V2`

PR #432, opened 2024-05-23
add new target EMAX_F405_ELRS_AIO

EmaxModel:
add new target EMAX_F405_ELRS_AIO，I use the previous the unified target to make a new config file, if there are any errors, please point out


nerdCopter:
Please do not PR from master, please P form a branch.  https://betaflight.com/docs/development#using-git-and-github
image

ot0tot:
Which unified target is this based on?

* Pin assignments should start with `P` and leading zeros should be omitted. e.g. `#define BEEPER_PIN PB4` not `#define BEEPER_PIN B04`. 
* Does this FC really have two hardware inverters?
* On-board mag and baro on SPI3?
* Flash and OSD should not be on the same SPI bus
* Why does the ESC_SERIAL pin have a timer assignment?

PR #474, opened 2024-06-24
add TUNERCF435

TUNERC-Aria:
Hello Betaflight Developers,

We provided samples of this AT32 FC before to the team before. A few changes were made on the official version of this FC. Detailed changes are listed below.

1. All of the connectors are moved to the bottom side of the FC and soldering pads are added to the top side of the FC.

2. Our company logo, model number, product name, orientation arrow and the soldering pad silkscreens are added to the official version.

3. The height of the power inductor is decreased from 3mm to 1.8mm.

4. The barometer model is changed from BMP280 to DPS310.
5. One LED is removed from the board (total number of the LED is changed from 3 to 2)

This FC will be put into mass production recently, we can provide production version samples to the developers if needed.

The first batch of samples is on the left, and the second batch is on the right.
noodle

### DIFF
--- a/configs/TUNERCF435/config.h
+++ b/configs/TUNERCF435/config.h
@@ -1,0 +1,102 @@
+/*
+ * This file is part of Betaflight.
+ *
+ * Betaflight is free software. You can redistribute this software
+ * and/or modify this software under the terms of the GNU General
+ * Public License as published by the Free Software Foundation,
+ * either version 3 of the License, or (at your option) any later
+ * version.
+ *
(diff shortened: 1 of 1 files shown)


ot0tot:
Flash and analog OSD should not be on the same SPI bus. Everything else looks OK.

TUNERC-Aria:
> Flash and analog OSD should not be on the same SPI bus. Everything else looks OK.

We need a very small package to replace the F411CEUx, which is relatively insufficient in performance, so we chose F435CGUx. The number of pins is limited, so we chose to put them on the same SPI bus.

PR #478, opened 2024-06-25
Update GEPRCF745_BT_HD

YI-BOYANG:
Some Baro and Mag were added, the default UART was configured, and the orientation of the No. 1 gyroscope was changed.


configs/GEPRCF745_BT_HD/config.h
```diff
@@ -35,6 +35,10 @@
 #define USE_BARO
 #define USE_BARO_BMP280
 #define USE_BARO_DPS310
+#define USE_BARO_MS5611
+#define USE_MAG
+#define USE_MAG_QMC5883
+#define USE_MAG_IST8310
```
haslinghuis, line 41: If compass is not physical onboard it should be user option.

```suggestion
#ifndef USE_MAG
#define USE_MAG
#define USE_MAG_QMC5883
#define USE_MAG_IST8310
#endif
```

haslinghuis (changes requested):

haslinghuis (approved):

nerdCopter:
`STM32F745`, should unified-target be updated as well?

nerdCopter (approved):

YI-BOYANG:
> `STM32F745`, should unified-target be updated as well?

Maybe not.

PR #493, opened 2024-07-10
TBS_LUCID_FC: update to latest board revision

bkleiner:
As discussed with @blckmn and @sugaarK, board needed some last minute revisions.

### DIFF
--- a/configs/TBS_LUCID_FC/config.h
+++ b/configs/TBS_LUCID_FC/config.h
@@ -55,13 +55,16 @@
 #define BEEPER_PIN                          PC13
 
 #define MOTOR1_PIN                          PC6
-#define MOTOR2_PIN                          PC7
+#define MOTOR2_PIN                          PB5
 #define MOTOR3_PIN                          PC8
 #define MOTOR4_PIN                          PC9
 #define MOTOR5_PIN                          PA0
 #define MOTOR6_PIN                          PA1
 #define MOTOR7_PIN                          PA2
 #define MOTOR8_PIN                          PA3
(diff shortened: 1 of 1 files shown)


haslinghuis:
Can we have PH0 / PH1 for I2C1 instead as to preserve using single GPIO port ?

configs/TBS_LUCID_FC/config.h
```diff
@@ -76,20 +79,20 @@
...
 #define UART7_TX_PIN                        PB4
 #define UART7_RX_PIN                        PB3
 
 #define UART8_TX_PIN                        PC2
 #define UART8_RX_PIN                        PC3
 
-#define I2C2_SCL_PIN                        PH2
-#define I2C2_SDA_PIN                        PH3
+#define I2C1_SCL_PIN                        PB6
+#define I2C1_SDA_PIN                        PC7
```
haslinghuis, line 95: Can we have PH0 / PH1 for I2C1 instead as to preserve using single GPIO port [for motors] - which also needs to be added to bus_i2c_atbsp_init.c?

bkleiner: PH0 / PH1 are the external crystal pins, we probably wanna preserve those.

haslinghuis (approved):

configs/TBS_LUCID_FC/config.h
haslinghuis: image

sugaarK (approved):

PR #507, opened 2024-07-27
Fix timers definition for FREEAXISF4 config.h.

ssoloshchak:
Fix: "No motor outputs found"
Update timers definition to be inline with config from unified-targets repo.

### DIFF
--- a/configs/FREEAXISF4/config.h
+++ b/configs/FREEAXISF4/config.h
@@ -68,10 +68,10 @@
 
 #define TIMER_PIN_MAPPING \
     TIMER_PIN_MAP( 0, PA3 , 3, -1) \
-    TIMER_PIN_MAP( 1, PB4 , 2,  0) \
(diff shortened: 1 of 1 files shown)


haslinghuis (approved):

ot0tot (approved):
Not sure why PA3, PA0, and PA2 need timer/dma defines. These can probably be removed.

ssoloshchak:
> Not sure why PA3, PA0, and PA2 need timer/dma defines. These can probably be removed.

To be honest I'm not mature enough in terms of configuration, just tried to fix no motors issue for my build :). Current version of code was tested in flight and I'd be happy to flash my build further with official build from BF configurator and not have to build changes locally again.

nerdCopter:
`#define GYRO_2_EXTI_PIN      NONE` necessary?

haslinghuis:
That would be the default anyway - so could have been removed :)

PR #514, opened 2024-08-05
fix PULLUP(s)

nerdCopter:
- fixes comoiling of these targets
- unsure if they should also have `#define USE_I2C_PULLUP`


nerdCopter:
so stupid of me, ran my git commands in wrong repo.

haslinghuis:
Agree this will need `#define USE_I2C_PULLUP` [instead] when we want it to be enabled.

PR #515, opened 2024-08-05
fix PULLUP(s)

nerdCopter:
- fixes compiling of these targets
- unsure if they should also have `#define USE_I2C_PULLUP`


configs/ALIENFLIGHTF4/config.h
```diff
@@ -120,4 +120,4 @@
 #define GYRO_1_SPI_INSTANCE SPI1
 #define GYRO_1_ALIGN CW270_DEG
 #define GYRO_1_ALIGN_YAW 2700
-#define I2C1_PULLUP ON
+//#define I2C1_PULLUP ON
```
haslinghuis, line 123: This needs to be
```suggestion
#define USE_I2C_PULLUP
```

ledvinap:
It seems that `USE_I2C_PULLUP` is used to enable pullups (on all ports). https://github.com/betaflight/betaflight/blob/master/src/main/target/common_defaults_post.h#L42-L52

haslinghuis:
All targets only use a single I2C port definition. If we want to have more control we need to change firmware instead.

nerdCopter:
whatever ya'll want. let me know how to modify this PR.

haslinghuis (approved):
If we fix it in firmware it will apply to firmware 4.5.2 / 4.6 only.

ot0tot:
Is this needed? The internal I2C pullups are very weak, I'm not sure enabling them does anything. There should be external pullups on the FC for I2C.

blckmn (approved):

PR #516, opened 2024-08-07
RRIOT_F722 - fix pinio to control VTX instead of PIT MODE

ws998116:
https://discord.com/channels/868013470023548938/1256215237879664672/1270436169309749370

We had pinio set up to control PIT MODE instead of the correct VTX control. This powers the VTX by default.

### DIFF
--- a/configs/RRIOT_F722/config.h
+++ b/configs/RRIOT_F722/config.h
@@ -99,5 +99,4 @@
 
 #define BEEPER_INVERTED
 
-#define PINIO1_PIN PB0
-#define PINIO1_BOX 39
+#define PINIO1_BOX 40


nerdCopter (approved):

configs/RRIOT_F722/config.h
```diff
@@ -99,5 +99,4 @@
 
 #define BEEPER_INVERTED
 
-#define PINIO1_PIN PB0
-#define PINIO1_BOX 39
+#define PINIO1_BOX 40
```
haslinghuis, line 102: This will not work without assigning a resource

haslinghuis (changes requested):

configs/RRIOT_F722/config.h
ws998116: This config allows us to power VTX by default. If users want to control it, they will need to use the CLI to assign the resource (we are providing instructions). We are doing this so that we do not need to solder the VTX jumper for the default setting we want. We have tested this and it works the way we need it to.

haslinghuis (approved):

PR #517, opened 2024-08-10
Add new target SPEEDYBEEF405V3S

SpeedyBee-OSS:
SPEEDYBEEF405V3S and SPEEDYBEEF405V3 are the same FC board, we changed BMI270 to ICM42688 in SPEEDYBEEF405V3S.
Due to ICM42688 align changed, we need add a new target to separate it.

### DIFF
--- a/configs/SPEEDYBEEF405V3S/config.h
+++ b/configs/SPEEDYBEEF405V3S/config.h
@@ -1,0 +1,131 @@
+/*
+ * This file is part of Betaflight.
+ *
+ * Betaflight is free software. You can redistribute this software
+ * and/or modify this software under the terms of the GNU General
+ * Public License as published by the Free Software Foundation,
+ * either version 3 of the License, or (at your option) any later
+ * version.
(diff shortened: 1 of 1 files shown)


ot0tot:
This target cannot be approved as already explained on the Discord. 

Due to the motor resource assignments chosen, this FC requires using burst or bitbang DMA to generate the motor outputs. In either of these modes, they gyro is not assigned DMA because of an issue with the F4 MCU. The only reason this kind of works on the F405V3 is because the BMI270 gyro has a 3.2kHz sample rate and it's possible to read in polled mode. The ICM42688p has an 8kHz sample rate, and cannot be read in polled mode without requiring significant CPU overhead. Combined with having blackbox and OSD on a shared SPI bus, this FC is a complete non-starter. 

This is the most poorly designed FC of the modern era, and a large part of the reason the Betaflight partners program exists is so this never happens again. Please discontinue this design and move on to the V4. This target will not be approved.

PR #565, opened 2024-09-09
Fix pinio and spi bus

dengdizzy:
Fix wrong io of pinio2,wrong gyro_1 and gyro_2 spi bus.

### DIFF
--- a/configs/TMOTORH743/config.h
+++ b/configs/TMOTORH743/config.h
@@ -93,7 +93,7 @@
 #define ADC_CURR_PIN PC3
 #define ADC_EXTERNAL1_PIN PC0
 #define PINIO1_PIN PC14
-#define PINIO2_PIN PD15
+#define PINIO2_PIN PD11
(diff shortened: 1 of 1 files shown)


haslinghuis (approved):

nerdCopter (approved):

ot0tot:
Are flash and OSD on the same SPI bus? Why is SPI4 defined if it is not used? Do you have a schematic of this FC we can review?

dengdizzy:
> Are flash and OSD on the same SPI bus? Why is SPI4 defined if it is not used? Do you have a schematic of this FC we can review?

Yes,flash and osd are using SPI3.SPI4 is not used on the board,but put SPI4 pads for external connection.This design for ardupilot firmware,SPI4 define can be deleted in betaflight config.

Here is the schematic,https://github.com/dengdizzy/TMOTOR-ELEC/blob/main/Betaflight/Pacer%20H743/Tmotor_H7_ONE-V2-5.pdf

PR #573, opened 2024-09-21
Remove USE_I2C_PULLUP

haslinghuis:


ledvinap:
is there rationale for this?

ot0tot:
I don't think this define makes sense, but it doesn't hurt anything. The internal pullups on these MCUs are very weak (~40kOhm), so it doesn't make a difference to I2C speeds whether or not they are enabled.

SteveCEvans:
Do we know if the FCs being affected have external pull-ups?

PR #586, opened 2024-10-09
Add ICM42688P to BETAFLIGHTF4

haslinghuis:
- See https://www.happymodel.cn/index.php/2024/01/05/super-f405hd-elrs-aio-3in1-flight-controller-built-in-uart-2-4g-elrs-and-20a-esc-for-hd-whoops-25-5x25-5mm-mount-hole/
- Should we remove MPU6000?


ot0tot:
Why are we allowing HappyModel to avoid the new target rules and reuse (and update) an existing target from another manufacturer?

The F405 has plenty of flash so adding additional defines isn't a problem. No reason to delete the MPU6000, which is used by the original manufacturer (FPV Model) of this FC.

PR #591, opened 2024-10-11
Add USE_WING define

ot0tot:
@limonspb Please test

### DIFF
--- a/configs/KAKUTEF4WING/config.h
+++ b/configs/KAKUTEF4WING/config.h
@@ -39,7 +39,9 @@
 #define USE_FLASH_W25Q128FV
 #define USE_MAX7456
 
+#ifndef USE_WING
 #define USE_WING
+#endif
 
 #define MOTOR1_PIN           PA8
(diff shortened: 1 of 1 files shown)


configs/KAKUTEF4WING/config.h
```diff
@@ -39,6 +39,8 @@
 #define USE_FLASH_W25Q128FV
 #define USE_MAX7456
 
+#define USE_WING
```
haslinghuis, line 42: Guess this will be a cloud build option - so we need a gate here.
```suggestion
#ifndef USE_WING
#define USE_WING
#endif
```

haslinghuis (approved):

limonspb:
How to use PRs from here for flashing?

haslinghuis:
@limonspb In this case simple add `WING` in custom defines until merged. Otherwise you would need to build locally and update local branch with changes made here.

nerdCopter (approved):

ledvinap (approved):

PR #598, opened 2024-10-22
Fix ORQAF7 - gate GHST RX protocol

haslinghuis:
- fixes https://build.betaflight.com/api/builds/6c8969734bf74fcc9f5baaee28c9e8d7/log


ot0tot:
Does this FC have a Ghost receiver built in? If so, should we put the serialrx define behind a #ifndef?

haslinghuis:
Put on draft as I'm awaiting answer from ORQA (discord) on how to proceed with integrated GSTH receiver.

ot0tot (approved):

PR #614, opened 2024-11-22
IFLIGHT_BORG_F435

sugaarK:


nerdCopter:
i changed the title and commit message

configs/ IFLIGHT_BORG_F435/config.h
```diff
@@ -0,0 +1,143 @@
...
+// BARO & MAG
+#define USE_BARO
+#define USE_BARO_BMP280
+#define USE_BARO_DPS310
+#define I2C1_SCL_PIN                    PB8
+#define I2C1_SDA_PIN                    PB9
+#define I2C2_SCL_PIN                    PH2
+#define I2C2_SDA_PIN                    PH3
+#define MAG_I2C_INSTANCE                (I2CDEV_1)
+#define BARO_I2C_INSTANCE               (I2CDEV_1)
```
haslinghuis, line 103:
```suggestion
#define MAG_I2C_INSTANCE                I2CDEV_1
#define BARO_I2C_INSTANCE               I2CDEV_1
```

sugaarK (commented):
nice

Jason-iFlight:
it’s tested already and it have been pass

configs/ IFLIGHT_BORG_F435/config.h
```diff
@@ -0,0 +1,143 @@
...
+#define PINIO1_PIN                      PC14
+
+// TIMERS & DMA
+#define TIMER_PIN_MAPPING               TIMER_PIN_MAP(  0, PA8,  1,  8 ) \
+                                        TIMER_PIN_MAP(  1, PA3,  1,  9 ) \
+                                        TIMER_PIN_MAP(  2, PB0,  2,  0 ) \
+                                        TIMER_PIN_MAP(  3, PB1,  2,  2 ) \
+                                        TIMER_PIN_MAP(  4, PC8,  2,  1 ) \
+                                        TIMER_PIN_MAP(  5, PC9,  2,  3 ) \
+                                        TIMER_PIN_MAP(  6, PB6,  1, 11 ) \
```
ot0tot, line 140: DMA request conflicts with ADC1

configs/ IFLIGHT_BORG_F435/config.h
```diff
@@ -0,0 +1,143 @@
...
+
+#pragma once
+
+#define FC_TARGET_MCU                   AT32F435G
+
+#define BOARD_NAME                      IFLIGHT_BORG_F435
+#define MANUFACTURER_ID                 IFRC
+
+// LEDS
+#define LED0_PIN                        PC15
```
ot0tot, line 30: PC15 cannot be used to drive a status LED

configs/ IFLIGHT_BORG_F435/config.h
```diff
@@ -0,0 +1,143 @@
...
+#define ADC_INSTANCE                    ADC1
+#define ADC1_DMA_OPT                    11
+#define DEFAULT_CURRENT_METER_SOURCE    CURRENT_METER_ADC
+#define DEFAULT_VOLTAGE_METER_SOURCE    VOLTAGE_METER_ADC
+
+// PINIO
+#define PINIO1_PIN                      PC14
+
+// TIMERS & DMA
+#define TIMER_PIN_MAPPING               TIMER_PIN_MAP(  0, PA8,  1,  8 ) \
```
ot0tot, line 134: Motor outputs are split over 4 timers, which is not acceptable. 
https://betaflight.com/docs/development/manufacturer/manufacturer-design-guidelines#32-resource-selection-considerations

configs/ IFLIGHT_BORG_F435/config.h
```diff
@@ -0,0 +1,143 @@
...
+                                        TIMER_PIN_MAP(  1, PA3,  1,  9 ) \
```
ot0tot, line 135: RX_PPM does not need DMA, and this timer conflicts with motors 7 & 8.
```suggestion
                                        TIMER_PIN_MAP(  1, PA3,  3,  -1 ) \
```

sugaarK:
@Jason-iFlight  needs some changes

configs/ IFLIGHT_BORG_F435/config.h
```diff
@@ -0,0 +1,143 @@
...
+// MOTORS & SERVOS
+#define MOTOR1_PIN                      PB0
+#define MOTOR2_PIN                      PB1
+#define MOTOR3_PIN                      PC8
+#define MOTOR4_PIN                      PC9
+#define MOTOR5_PIN                      PB6
+#define MOTOR6_PIN                      PB7
+#define MOTOR7_PIN                      PB11
+#define MOTOR8_PIN                      PB10
+#define RX_PPM_PIN                      PA3
```
ot0tot, line 45:
```suggestion

```

configs/ IFLIGHT_BORG_F435/config.h
```diff
@@ -0,0 +1,143 @@
...
+#define USE_BARO
+#define USE_BARO_BMP280
+#define USE_BARO_DPS310
+#define I2C1_SCL_PIN                    PB8
+#define I2C1_SDA_PIN                    PB9
+#define I2C2_SCL_PIN                    PH2
+#define I2C2_SDA_PIN                    PH3
+#define MAG_I2C_INSTANCE                I2CDEV_1
+#define BARO_I2C_INSTANCE               I2CDEV_1
+#define DEFAULT_BARO_DPS310
```
ot0tot, line 104:
```suggestion

```

configs/ IFLIGHT_BORG_F435/config.h
```diff
@@ -0,0 +1,143 @@
...
+#define UART3_RX_PIN                    PC11
+#define UART3_TX_PIN                    PC10
+#define UART4_RX_PIN                    PA1
+#define UART4_TX_PIN                    PA0
+#define UART5_RX_PIN                    PD2
+#define UART5_TX_PIN                    PC12
+#define UART6_RX_PIN                    PC7
+#define UART6_TX_PIN                    PC6
+#define USB_DETECT_PIN                  PC5
+#define SERIALRX_UART                   SERIAL_PORT_USART2
```
ot0tot, line 120: UART2 is susceptible to DFU hijacking and should not be used for receivers/GPS

configs/ IFLIGHT_BORG_F435/config.h
```diff
@@ -0,0 +1,143 @@
...
+
+// ADC
+#define ADC_VBAT_PIN                    PC2
+#define ADC_CURR_PIN                    PC1
+#define ADC_INSTANCE                    ADC1
+#define ADC1_DMA_OPT                    11
```
ot0tot, line 126:
```suggestion
#define ADC1_DMA_OPT                    14
```

sugaarK:
@Jason-iFlight  this will need retesting and PROPER testing... you would have had issues with the conflicts...

sugaarK:
can you address any of the issues?

configs/ IFLIGHT_BORG_F435/config.h
Jason-iFlight: it is said that this ADC1 uses DMA to cause conflict, but ADC-CURR does not use this function due to inaccurate data, and I will disable this function.

configs/ IFLIGHT_BORG_F435/config.h
Jason-iFlight: if we do not consider the 4 motors of the motor 5~8, only the 4 motors of 1~4 can be used, is it that these conflicts do not exist?

configs/ IFLIGHT_BORG_F435/config.h
Jason-iFlight: uart2 will not use on receiver and GPS，However, UART2 is used as power callback, power callback only works when the battery is powered on, there is no conflict in entering DFU, because entering DFU is usually USB powered state, and the battery is powered off.

configs/ IFLIGHT_BORG_F435/config.h
sugaarK: so we removing adc1 define

configs/ IFLIGHT_BORG_F435/config.h
sugaarK: we removing motors 5-8?

nerdCopter:
- rebased it on master, and force-pushed.
- fixed the space in the folder-name.

nerdCopter (approved):
- approving for workflow only.
- did not deeply review, but did read some conversations
- defer to other devs for validating the config.

sugaarK:
@ot0tot this thing is in production.. they understand it will not be accepted to bring us a target like this already in production next time or they will be ignored.. can you commit any changes we can make in firmware and then we will approve as is.. this is the final time of this though next time we will simply reject a target like this.

configs/ IFLIGHT_BORG_F435/config.h
ot0tot: > it is said that this ADC1 uses DMA to cause conflict, but ADC-CURR does not use this function due to inaccurate data, and I will disable this function.

What does this mean?

sugaarK: @Jason-iFlight ?

configs/ IFLIGHT_BORG_F435/config.h
sugaarK: @Jason-iFlight ?

configs/ IFLIGHT_BORG_F435/config.h
Jason-iFlight: it mean we can move this one out.

configs/ IFLIGHT_BORG_F435/config.h
Jason-iFlight: yeah

configs/ IFLIGHT_BORG_F435/config.h
Jason-iFlight: we will not use it for receivers/GPS

configs/ IFLIGHT_BORG_F435/config.h
ot0tot: Move what one of of where?

configs/ IFLIGHT_BORG_F435/config.h
ot0tot: Can you show the wiring guide for this FC?

ot0tot:
This seems to be the same as the old https://github.com/betaflight/config/blob/master/configs/IFLIGHT_BLITZ_F435/config.h target. Why do we need to add this also?

If they're not willing to correct any of the design issues, might as well just use the old grandfathered target.

configs/ IFLIGHT_BORG_F435/config.h
sugaarK: @Jason-iFlight you have been asked questions... HERE....

configs/ IFLIGHT_BORG_F435/config.h
sugaarK: @Jason-iFlight HERE

sugaarK:
> This seems to be the same as the old https://github.com/betaflight/config/blob/master/configs/IFLIGHT_BLITZ_F435/config.h target. Why do we need to add this also?
> 
> If they're not willing to correct any of the design issues, might as well just use the old grandfathered target.

would be easier and less hassle

PR #630, opened 2024-12-05
update StellarF4 target

cvetaevvitaliy:
We have a lot of problems with the BMI270 gyroscope, all new FCs will have icm42688p


haslinghuis (approved):
Does the updated IMU implement a dedicated LDO ?

sugaarK:
> Does the updated IMU implement a dedicated LDO ?

good question.. you'd be better with the BMI if it doesn't have a seperate LDO

cvetaevvitaliy:
> Does the updated IMU implement a dedicated LDO ?

yes, IMU has separate LDO

configs/STELLARF4/config.h
```diff
@@ -25,12 +25,17 @@
 #define BOARD_NAME STELLARF4
 #define MANUFACTURER_ID STBE
 
+#define TARGET_BOARD_IDENTIFIER "STF4"
+#define USBD_PRODUCT_STRING     "StingBee Stellar F4"
```
nerdCopter, line 29: nothing wrong with this, just noting that `src/platform/STM32/target/STM32F405/target.h` contains
```
#ifndef USBD_PRODUCT_STRING
#define USBD_PRODUCT_STRING     "Betaflight STM32F405"
#endif
```

nerdCopter (approved):

ot0tot (approved):

configs/STELLARF4/config.h
haslinghuis: config.h is processed before target.h

PR #634, opened 2024-12-07
Add DAKEFPVF435

ThanYangFPV:
ADD DAKEFPVF435


configs/DAKEFPVF435/config.h
```diff
@@ -0,0 +1,130 @@
...
+#define PINIO2_CONFIG 129
+#define PINIO2_BOX 41
+
+#define GYRO_1_SPI_INSTANCE SPI1
+#define GYRO_1_ALIGN CW90_DEG
+#define GYRO_1_ALIGN_YAW 900
+#define DEFAULT_CURRENT_METER_SCALE 120
+
+#define SERIALRX_UART           SERIAL_PORT_UART5
+#define ESC_SENSOR_UART         SERIAL_PORT_USART3
```
haslinghuis, line 130: No new line at EOF
```suggestion
#define ESC_SENSOR_UART         SERIAL_PORT_USART3

```

configs/DAKEFPVF435/config.h
```diff
@@ -39,9 +39,7 @@
 #define USE_FLASH
 #define USE_FLASH_M25P16
 #define USE_MAX7456
-#ifndef USE_CAMERA_CONTROL
-#define USE_CAMERA_CONTROL
-#endif
+#define USE_VTX
```
haslinghuis, line 42: USE_VTX is a build option
```suggestion
```

haslinghuis (approved):

configs/DAKEFPVF435/config.h
```diff
@@ -0,0 +1,127 @@
...
+#define TIMER_PIN_MAPPING          \
+    TIMER_PIN_MAP( 0, PA0 , 1,  0) \
+    TIMER_PIN_MAP( 1, PA1 , 1,  1) \
+    TIMER_PIN_MAP( 2, PA2 , 1,  2) \
+    TIMER_PIN_MAP( 3, PA3 , 1,  3) \
+    TIMER_PIN_MAP( 4, PA8 , 1,  4) \
+    TIMER_PIN_MAP( 5, PA9 , 1,  5) \
+    TIMER_PIN_MAP( 6, PA10, 1,  6) \
+    TIMER_PIN_MAP( 7, PC9 , 2,  7) \
+    TIMER_PIN_MAP( 8, PC8 , 2,  8) \
```
ot0tot, line 96: LED_STRIP cannot be on the same timer as M8
```suggestion
    TIMER_PIN_MAP( 8, PC8 , 1,  8) \
```

configs/DAKEFPVF435/config.h
```diff
@@ -0,0 +1,127 @@
...
+#define BEEPER_INVERTED
+#define PINIO1_CONFIG 129
+#define PINIO1_BOX 40
+#define PINIO2_CONFIG 129
+#define PINIO2_BOX 41
+
+#define GYRO_1_SPI_INSTANCE SPI1
+#define GYRO_1_ALIGN CW90_DEG
+#define GYRO_1_ALIGN_YAW 900
+#define DEFAULT_CURRENT_METER_SCALE 120
```
ot0tot, line 124: Current meter scale should not be defined unless this is an AIO or only comes as a stack, as it depends entirely on the ESC.
```suggestion

```

haslinghuis: This is used in 320 targets already :confused:

ThanYangFPV: The value of this current is a match based on our stack.

configs/DAKEFPVF435/config.h
```diff
@@ -0,0 +1,126 @@
...
+
+#define BEEPER_INVERTED
+#define PINIO1_CONFIG 129
+#define PINIO1_BOX 40
+#define PINIO2_CONFIG 129
+#define PINIO2_BOX 41
+
+#define GYRO_1_SPI_INSTANCE SPI1
+#define GYRO_1_ALIGN CW90_DEG
+#define GYRO_1_ALIGN_YAW 900
```
ot0tot, line 123: Shouldn't this just be CW180_DEG?
```suggestion
#define GYRO_1_ALIGN CW180_DEG

```

ThanYangFPV: image
The gyroscope on our PCB is mounted in this direction, and we have always been like this, so there should be no problem at all, right?

ot0tot: You are saying the gyro is rotated 90 degrees, then 90 degrees again. Why not just say 180?

ThanYangFPV: Yes, you're right!

ot0tot (approved):

PR #644, opened 2024-12-19
Fix mutual exclusive gyro alignment

haslinghuis:
See

- https://github.com/betaflight/config/pull/643
- https://github.com/betaflight/betaflight/pull/14092
- https://github.com/betaflight/config/pull/645 for improvement


configs/ACCIF405/config.h
```diff
@@ -60,7 +60,6 @@
 #define GYRO_1_CS_PIN        PA4
 #define GYRO_1_SPI_INSTANCE  SPI1
 #define GYRO_1_ALIGN CW0_DEG
-#define GYRO_1_ALIGN_YAW 1800
```
ledvinap, line 63: This one is wrong - only remove cases where GYRO_1_ALIGN and GYRO_1_ALIGN_YAW do match. These mismatches must be handled individually (or left to trigger error)

configs/AIRBOTG4AIO/config.h
```diff
@@ -120,5 +120,4 @@
 
 #define GYRO_1_SPI_INSTANCE          SPI1
 #define GYRO_1_ALIGN                 CW0_DEG_FLIP
```
ledvinap, line 122: total mismatch

configs/AOCODAF405/config.h
```diff
@@ -106,4 +106,3 @@
 #define FLASH_SPI_INSTANCE SPI3
 #define GYRO_1_SPI_INSTANCE SPI1
 #define GYRO_1_ALIGN CW0_DEG_FLIP
```
ledvinap, line 108: !

configs/CHONKERH735/config.h
```diff
@@ -123,8 +123,6 @@ Details: https://github.com/crteensy/yolo-chonker/tree/as-built-20230303
 
 #define DEFAULT_GYRO_TO_USE             GYRO_CONFIG_USE_GYRO_1
 #define GYRO_1_ALIGN                    CW0_DEG_FLIP
-#define GYRO_1_ALIGN_PITCH              1800
```
ledvinap, line 126: (this one is correct - 0deg + flip)

configs/SPRACINGH7EXTREME/config.h
```diff
@@ -245,6 +245,3 @@
 #define GYRO_1_ALIGN CW180_DEG
 #define GYRO_2_SPI_INSTANCE SPI2
 #define GYRO_2_ALIGN ALIGN_CUSTOM
-#define GYRO_2_ALIGN_ROLL 0
-#define GYRO_2_ALIGN_PITCH 0
-#define GYRO_2_ALIGN_YAW 2250
```
ledvinap, line 250: THis one must stay - it is intentional CUSTOM alignment

configs/STELLARF7V2/config.h
```diff
@@ -118,7 +118,6 @@
 #define FLASH_SPI_INSTANCE SPI3
 #define GYRO_1_SPI_INSTANCE SPI1
 #define GYRO_1_ALIGN CW180_DEG
-#define GYRO_1_ALIGN_PITCH 1800
```
ledvinap, line 121: Wrong

configs/STELLARH7V2/config.h
```diff
@@ -146,7 +146,6 @@
 #define PINIO1_BOX 40
 #define GYRO_1_SPI_INSTANCE SPI1
 #define GYRO_1_ALIGN CW180_DEG
-#define GYRO_1_ALIGN_PITCH 1800
```
ledvinap, line 149: !

ledvinap (commented):
Marked some problems, but not all.
My PR removes trivial cases, I can improve the script to handle FLIP too. Then remaining CUSTOM align needs individual review/fix

haslinghuis:
- Rather fixing here would like #645 to handle exceptions.

configs/STELLARF7V2/config.h
haslinghuis: Here I'm confused why this is valid.

configs/CHONKERH735/config.h
haslinghuis: Would think to remove `#define GYRO_1_ALIGN                    CW0_DEG_FLIP` instead.

configs/STELLARF7V2/config.h
ledvinap: 180 pitch is cw0_flip

haslinghuis: Sorry for ignorance - as you stated settings are mutual exclusive - in my understanding when using `GYRO_1_ALIGN CW180_DEG` custom definition should be unavailable

image

ledvinap: I'm talking about config.h - it should use either GYRO_1_ALIGN CW180_DEG or #define GYRO_1_ALIGN_YAW 1800. In this case, ALIGN is CW180_DEG but custom alignment is GYRO_1_ALIGN_PITCH 1800. This does not match, there is some problem with config.

PR #648, opened 2024-12-20
Fix custom alignment

haslinghuis:
See comment https://github.com/betaflight/config/pull/645#issuecomment-2556927473


nerdCopter (approved):

ledvinap:
Why? Valid value is -3600..3600.

haslinghuis:
@ledvinap

IMHO if signed it should be -1800..1800 and why use decidegrees. Only clockwise may be easier but just a preference.
Should I change to `-450` instead ?

ledvinap:
I'd keep current state for now, but this PR won't hurt ...

ledvinap (approved):
Decide yourself ..

PR #649, opened 2024-12-22
Remove legacy ACCGYRO driver from STM32F411DISCOVERY

haslinghuis:
- removing legacy accgyro drivers in preparation for https://github.com/betaflight/betaflight/pull/14087


blckmn:
I think if prefer to keep the driver...

nerdCopter (approved):

blckmn (changes requested):
Rather than remove this gyro, acc combo why not modify them to be non legacy.

haslinghuis:
Thinking we should just purge this configuration instead - as we no longer recommend F411 designs.

blckmn:
Whilst that true... the F411 disco is still a good debugging platform for general code checking. When we stop supporting F411 entirely it'd make sense to drop it.

PR #673, opened 2025-01-21
Remove custom gyro defines when align preset is used

haslinghuis:
- As most configs were migrated from unified targets - cli contains values for gyro align preset and custom alignment results. 
- Removing custom alignment where preset is present


nerdCopter (approved):
- approving untested

ledvinap (approved):
IIRC GYRO_n_ALIGN did override rpy, so this should be pretty safe. 
but i don't understand what caused this mismatch.

PR #685, opened 2025-02-05
LUXMINIF7 Gyro change

freasy:
The new LUXMINIF7 PCB has the ICM gyro, while the older versions have MPU6k.


nerdCopter:
@freasy , this page show a version with ICM20602, so i believe the 6500 should be retained.
https://www.getfpv.com/lumenier-lux-mini-f7-20x20-stack.html

PR #690, opened 2025-02-10
JHEF745V2 -- Disable dshot bitbang

haslinghuis:
Fixes the issue when using motor 5 it locks up firmware


haslinghuis:
@ot0tot please review

ot0tot:
Disabling DShot bitbang has a number of detrimental effects on DMA assignments when more than 4 motors are used. Is there an issue thread or support ID from this FC with bitbang enabled? It is unfortunate that the motors are spread over 4 ports, but with timers 1 and 8 both open I would still expect this to work. 

I see one error with the timer map, where TIM1_CH1 should be on DMA2 Stream6 Channel 0 so it does not conflict with SPI4_RX. Not sure if that will resolve the issue with 5+ motors. Can someone test?

ot0tot (approved):

PR #708, opened 2025-03-05
Fix gyro alignment on JHEG474

ot0tot:
Gyro alignment was wrong on my FC. Can someone else verify it should be CW90?


nerdCopter (approved):
- approving for workflow; untested.

haslinghuis (approved):
CW90 is correct.

PR #723, opened 2025-03-17
BOTWINGF405 HSE

ot0tot:
Confirm crystal freq. and add HSE define.

### DIFF
--- a/configs/BOTWINGF405/config.h
+++ b/configs/BOTWINGF405/config.h
@@ -117,7 +117,7 @@
 #define ADC_INSTANCE                        ADC1
 
 #define DEFAULT_BLACKBOX_DEVICE             BLACKBOX_DEVICE_FLASH
-#define SYSTEM_HSE_MHZ                      8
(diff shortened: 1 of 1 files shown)


configs/BOTWINGF405/config.h
```diff
@@ -117,6 +117,7 @@
 #define ADC_INSTANCE                        ADC1
 
 #define DEFAULT_BLACKBOX_DEVICE             BLACKBOX_DEVICE_FLASH
+#define SYSTEM_HSE_MHZ                      8
```
SteveCEvans, line 120: This should be 24.

SteveCEvans (changes requested):

configs/BOTWINGF405/config.h
haslinghuis, line 120:
```suggestion
#define SYSTEM_HSE_MHZ                      24
```

haslinghuis (approved):
Sample should arrive on Monday

SteveCEvans (approved):

PR #727, opened 2025-03-19
Add CRAZYBEE473 target

ot0tot:


haslinghuis (approved):

ot0tot:
The difference between 110 and 113 is very small, and depends on the tolerance of the resistors used for the Vbat voltage divider. Let's check a few more FCs and see what the average value is.

nerdCopter (approved):

ot0tot:
My FC tests with a voltage scale closer to 135, but the response is non-linear and seems quite slow. Will follow-up in the manufacturer channel.

sugaarK:
> @ot0tot Hey man , please help to adjust voltage divider to 113 as @sugaarK said. Thanks!

Did you check a few others? It was way off on my example

ot0tot:
I only have one FC.

PR #734, opened 2025-03-27
Enable dshot bitbang on JHEF745V2

haslinghuis:
When bidirectional dshot is used
- fixes battery not initializing 
- fixes errors on motors 3 and 4

NOTE: To use 8 motors need to disable bidirectional dshot and dshot bitbang before switching model


nerdCopter (approved):

ot0tot:
Just mentioning this PR for posterity 😅 
https://github.com/betaflight/config/pull/690#issuecomment-2683725858

ot0tot (approved):
Approving, but we should also remove the additional motor defines that are causing conflicts. 
https://github.com/betaflight/config/pull/690

haslinghuis:
@ot0tot using 8 motors is only possible when `dshot bitbang` and `dshot telemetry` are both disabled.

Current config now works for quad out the box. More experienced users should adjust settings.

PR #747, opened 2025-04-09
Update ADC

haslinghuis:
@crteensy see comment in https://github.com/betaflight/betaflight/blob/609533a73704f9c9173d57af0c85cd403272d778/src/platform/STM32/adc_stm32h7xx.c#L216


nerdCopter (approved):
- approving untested

ot0tot (approved):
Not necessary to remove 
```
#define ADC1_DMA_OPT                    8
#define ADC3_DMA_OPT                    9
```
but this is fine.

PR #749, opened 2025-04-11
Add DPS310 to WARPF7

haslinghuis:


sugaarK:
isn't the DSP310 EOL with only poor performing clones now available

nerdCopter (approved):

ot0tot (approved):

ot0tot:
DPS310 driver is also used for the newer SPL07_003

PR #764, opened 2025-04-25
DAKEFPV support 12S batteries

ThanYangFPV:
In order to support 12S batteries, we changed the voltage divider resistor of VBAT from 10K/100K to 10K/150K, and we want to change the target DEFAULT_VOLTAGE_METER_SCALE from the default 110 to 160.


ot0tot:
Have you shipped any of these already with the old voltage divider ratio?

Current sensor scale should not be set unless the ESC is built-in (AIO) or the FC is sold as part of a stack.

DSHOT300 does not need to be set in the config.

ThanYangFPV:
> Have you shipped any of these already with the old voltage divider ratio?
> 
> Current sensor scale should not be set unless the ESC is built-in (AIO) or the FC is sold as part of a stack.
> 
> DSHOT300 does not need to be set in the config.

The DAKEF405 and DAKEF722 had been dispatched already. The customer agrees to continue using the 8S battery. But, the customer requests that all subsequent shipments support up to 12S. The F435, F722X8, and H743 have not been dispatched yet. The customer also requests that all of them support 12S.

We do AIO and sell it separately as an accessory.

DSHOT300 You can leave it out of config, no problem.

ThanYangFPV:
@ot0tot

PR #767, opened 2025-05-03
add disable gyro clock

cvetaevvitaliy:
add disable gyro clock for testing and comparing results ICM42688P

Usage
* add `-DUSE_DISABLE_GYRO_CLKIN` using `EXTRA_FLAGS` for local build
* add  `DISABLE_GYRO_CLKIN` using `custom define` for cloud build

### DIFF
--- a/configs/STELLARH7DEV/config.h
+++ b/configs/STELLARH7DEV/config.h
@@ -183,6 +183,11 @@
 #define GYRO_2_CLKIN_PIN                PB4
 #endif
 
+// Disable GYRO_CLKIN for testing and comparing results
+#ifdef USE_DISABLE_GYRO_CLKIN
+#undef USE_GYRO_CLKIN
+#endif
+
 // Blackbox
 #define USE_FLASH
 #define USE_FLASH_W25M02G


haslinghuis (approved):

ot0tot:
Why not just disable the resource in the CLI?

cvetaevvitaliy:
It is much more convenient to disable before build


сб, 3 трав. 2025 р. о 23:20 ot0tot ***@***.***> пише:

SteveCEvans:
> Why not just disable the resource in the CLI?

Exactly my thought. That’s easy and obvious, whereas build options are not so easily remembered and have a habit of becoming broken over time when the code is updated and their significance is overlooked.

cvetaevvitaliy:
> > Why not just disable the resource in the CLI?
> 
> 
> 
> Exactly my thought. That’s easy and obvious, whereas build options are not so easily remembered and have a habit of becoming broken over time when the code is updated and their significance is overlooked.


Only non-developer wouldn't do this. Finally it's only for one target and as part of testing this feature for a quick comparison - Enable/Disable gyro clock

ot0tot:
If only for testing it makes even more sense to just do it in the CLI. No need to recompile the firmware and reflash.

SteveCEvans:
Exactly. Once tested and proven it would be dead, unused complexity.

haslinghuis:
Closing as developers can just disable the pin (no need for new local define)

PR #770, opened 2025-05-07
Fix redundant port resource on BOTWINGF722

haslinghuis:


configs/BOTWINGF722/config.h
```diff
@@ -126,8 +126,6 @@
 
 #define RX_PPM_PIN                          PA3
 
-#define UART4_RX_PIN                        PA1
-
 #define UART6_TX_PIN                        PC6
 #define UART6_RX_PIN                        PC7
```
osirisinferi, line 130: Looks like these UART6 defines are also redundant with lines 124 and 125?

ot0tot (approved):

PR #775, opened 2025-05-13
added Aerium Radian F405 flight controller

yaadch:
Added Aerium Radian F405 HW definitions
https://www.aerium.co.il/product-page/radian

image

### DIFF
--- a/configs/AERIUM_RADIAN_F405/config.h
+++ b/configs/AERIUM_RADIAN_F405/config.h
@@ -1,0 +1,115 @@
+/*
+ * This file is part of Betaflight.
+ *
+ * Betaflight is free software. You can redistribute this software
+ * and/or modify this software under the terms of the GNU General
+ * Public License as published by the Free Software Foundation,
+ * either version 3 of the License, or (at your option) any later
+ * version.
+ *
+ * Betaflight is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ *
+ * See the GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public
+ * License along with this software.
+ *
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#define FC_TARGET_MCU     STM32F405
+
+#define BOARD_NAME        AERIUM_RADIAN_F405
+#define MANUFACTURER_ID   AERIUM
+
+#define USE_GYRO
+#define USE_SPI_GYRO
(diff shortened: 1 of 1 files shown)


haslinghuis:
Please check https://betaflight.com/docs/development/manufacturer/config-target-guidance

ot0tot:
@sugaarK Have you received this schematic?

ot0tot:
@yaadch Please review the manufacturer design guidelines (https://betaflight.com/docs/development/manufacturer/manufacturer-design-guidelines) and Betaflight Connector Standard (https://betaflight.com/docs/development/manufacturer/connector-standard). This design has many issues and cannot be approved as-is.

sugaarK:
> > From our perspective, after reviewing the requirements multiple times, we haven’t identified the extent of the issues being raised.
> 
> I don't know what issues the BF team has found, but on the page linked in the first reply it says:
> 
> > From the 3rd of December 2024, the Betaflight project will no longer accept new flight controller designs using the STM F4 and F7 series microcontroller where there is more than 4 motor outputs.
> 
> I notice a F405 board with 8 motor pins defined.
> 
> That's just what caught my eye, there might be more.

any pre opened before the ban were allowed to continue.. its not negotiable frankly if This is meant to be a premium product use an h7. the price difference is no longer an issue and you'll have a far more capable and flexible flight controller.  the reason for the motor out put limit is its that or we just stop using the restricted chipsets..

configs/AERIUM_RADIAN_F405/config.h
```diff
@@ -0,0 +1,114 @@
...
+ *
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#define FC_TARGET_MCU     STM32F405
+
+#define BOARD_NAME        AERIUM_RADIAN_F405
+#define MANUFACTURER_ID   AERIUM
```
sugaarK, line 27: 4 letter code is used here.

configs/AERIUM_RADIAN_F405/config.h
```diff
@@ -0,0 +1,114 @@
...
+#define MAG_I2C_INSTANCE     I2CDEV_1
+#define MAG_I2C_ADDRESS      28
+
+#define BEEPER_PIN           PA0
+
+#define MOTOR1_PIN           PB1
+#define MOTOR2_PIN           PB0
+#define MOTOR3_PIN           PA8
+#define MOTOR4_PIN           PA3
+#define MOTOR5_PIN           PA2
```
sugaarK, line 69: this chipset is limited to 4 motor outputs and timer only bidirectional  dshot. also these motors must be pinned in a way to ensure they use only one timer for the motor set..

sugaarK (changes requested):

Issue #778, opened 2025-05-14
Sub250 Nimble65 (GEPRCF411SX1280/STM32F411) unflyable on Betaflight 4.5.2 with current config due to gyro misconfiguration (4.5.1 okay)


DavidAnson:
The Sub250 Nimble65 ships (as of about 1 month ago) with <https://build.betaflight.com/api/builds/c6e8ae1147c3a84dcf4095780dd14204/log> flashed containing Betaflight 4.5.1 with config `ad0057e`. That flies fine.

The corresponding Cloud Build for Betaflight 4.5.2 is <https://build.betaflight.com/api/builds/474ed9cd32f0201161e0510e5374e625/log> with config `467f87b`. It idles okay, but if you try to take off, ZOOM TO THE MOON with audible power fluctuations on the way up. Upon inspection, the gyro is misaligned and this can be easily observed in Betaflight Configurator's Setup tab - it's about 45 degrees off so pure physical pitch or roll produce measured pitch AND roll on screen.

If you're lucky, you notice that after flashing. If not, you may have to disarm at 30 feet over concrete and crack your frame in 3 places. :|

If you're the kind of person who likes to `diff all`, you may notice that in the working configuration immediately after flashing:
```
set gyro_1_sensor_align = CW0
set gyro_1_align_yaw = 0
```

And in the broken configuration immediately after flashing:
```
set gyro_1_sensor_align = CUSTOM
set gyro_1_align_yaw = 3150
```

If you're the kind of person who likes to read GitHub commits, you may decide the problem is a COMBINATION of these two commits:

Betaflight 4.5.2: <https://github.com/betaflight/betaflight/commit/7e55f9e73c9628c7fd4309bc6039495f13db9f1d>
Betaflight config: <https://github.com/betaflight/config/commit/ac7c65c3acfef6b60c1b70fce64a381cc958feba#diff-f8cb1e76e2e76dd5c148e83ff603d845ef5d004dfa8964d3a595dbaf9ee88c20>, specifically `configs/GEPRCF411SX1280/config.h`, extra specifically:
```
+ #define GYRO_1_ALIGN ALIGN_CUSTOM
#define GYRO_1_ALIGN_YAW -45
```

I don't know enough about `GEPRCF411SX1280` to say if this configuration is wrong for all such boards, but I have confirmed experimentally that it is wrong for the Sub250 Nimble65. The fix (also confirmed experimentally) is to apply the two "working configuration" settings above after flashing Betaflight 4.5.2.

[...]

PR #779, opened 2025-05-14
Fix alignment on GEPRCF411SX1280

haslinghuis:
- fixes: #778


ot0tot:
Issue is for a different quad from a different manufacturer, and has nothing to do with the GEPRC target.

haslinghuis:
No support for clones !

haslinghuis:
Needs testing (made this up after comparing unified target with config)

ot0tot:
The existing config is correct. No changes needed.

PR #781, opened 2025-05-15
Align voltage meter with real measurements for FLYTEX405V2

MatviiG:
This patch changes voltage meter and current meter settings to values which allows correct measurement of current and voltage. Also board ID and PID for USB connection added.

## Pull-Request requirements

**Mandatory Review for All New Flight Controllers**

- All new flight controllers must undergo the Betaflight review process, regardless of whether they use an existing target.
- Manufacturers may reuse the same target for multiple designs, but each new hardware release must be reviewed before approval.

**Hardware Compliance Requirements**

- All hardware must adhere to the [Manufacturer Design Guidelines](https://betaflight.com/docs/development/manufacturer/manufacturer-design-guidelines).
- Poor hardware designs will not be accepted, as outlined in the [Requirements for the Submission of New and Updated Targets](https://betaflight.com/docs/development/manufacturer/requirements-for-submission-of-targets).
- Adhere to the [Config Target Guidance] https://betaflight.com/docs/development/manufacturer/config-target-guidance
- To avoid costly issues and delays, we strongly encourage consulting with Betaflight developers early in your design process and providing schematics for review **before** producing hardware.

These measures help maintain high standards and ensure compatibility within the Betaflight ecosystem.

If you have any questions or need guidance, feel free to reach out to the Betaflight development team.

**Housekeeping** 
- Pull-Request only from a custom branch, not `master`.
- Replace this text with details of your own.


configs/FLYTEX405V2/config.h
```diff
@@ -106,8 +109,9 @@
 #define DEFAULT_DSHOT_BITBANG DSHOT_BITBANG_OFF
 #define DEFAULT_CURRENT_METER_SOURCE CURRENT_METER_ADC
 #define DEFAULT_VOLTAGE_METER_SOURCE VOLTAGE_METER_ADC
-#define DEFAULT_VOLTAGE_METER_SCALE 110
-#define DEFAULT_CURRENT_METER_SCALE 326
+#define DEFAULT_VOLTAGE_METER_SCALE 105
+#define DEFAULT_VOLTAGE_METER_DIVIDER 5
+#define DEFAULT_CURRENT_METER_SCALE 200
```
ot0tot, line 114: These values don't make sense. Based on the provided schematic the scale should be 190. Current meter scale should not be set in the config unless the FC is an AIO or sold as part of a stack. 
```suggestion
#define DEFAULT_VOLTAGE_METER_SCALE 190
#define DEFAULT_CURRENT_METER_SCALE 200
```

MatviiG: we would like to change divider R3/R6 to bigger value ( 200/10 ) as we are going to use 8S battery. I understand that it is only 11% increase but anyway we have to change initially incorrect value.

configs/FLYTEX405V2/config.h
```diff
@@ -106,8 +109,8 @@
 #define DEFAULT_DSHOT_BITBANG DSHOT_BITBANG_OFF
 #define DEFAULT_CURRENT_METER_SOURCE CURRENT_METER_ADC
 #define DEFAULT_VOLTAGE_METER_SOURCE VOLTAGE_METER_ADC
-#define DEFAULT_VOLTAGE_METER_SCALE 110
-#define DEFAULT_CURRENT_METER_SCALE 326
+#define DEFAULT_VOLTAGE_METER_SCALE 190
```
MatviiG, line 112:
```suggestion
#define DEFAULT_VOLTAGE_METER_SCALE 210
```
in case 200/10 divider it should be proper value

haslinghuis (approved):

nerdCopter (approved):

PR #799, opened 2025-06-03
Target/hummingbird aio255 atf435

LYNHQQ:
Add the HUMMINGBIRD AIO255 AT32F435 target.

## Summary by CodeRabbit

- **Chores**
  - Removed configuration files for a wide range of flight controller boards. These files previously defined hardware-specific settings, sensor and peripheral support, and pin assignments for each board. As a result, firmware builds for these specific boards are no longer supported. No user-facing features or bug fixes are included in this change.


osirisinferi:
@LYNHQQ Also as with #800, this PR removes *all* existing configurations. Please rebase to upstream master instead of your own `master` branch which is clearly not fit for branching off from.

PR #838, opened 2025-07-18
add new sensors to target for Hobbywing

sugaarK:
## Summary by CodeRabbit

* **New Features**
  * Expanded hardware sensor support to include the ICM42688P accelerometer and gyroscope, as well as the DPS310 barometer, for the Hobbywing XRotor F7 Conv board.


KarateBrot (changes requested):
You need to rename the actual target in the config file, too.

haslinghuis (changes requested):
- name change does not match other hobbywing boards
- in general would not work with auto-detect for existing users resulting in additional support requests.

This board is on the market at least for two years.

sugaarK:
have reverted name and added the newer sensors to the target

haslinghuis (approved):

haslinghuis (approved):

KarateBrot (approved):

haslinghuis (dismissed):
Are schematics verified to use required LDO for ICM42688P ?

nerdCopter (approved):

PR #866, opened 2025-08-18
SPEEDYBEEF405V5: Adjust current meter offset and scale

kedeng:
Adjust current meter offset and scale.

## Summary by CodeRabbit

- New Features
  - Added default current meter calibration for the SPEEDYBEEF405V5 target, providing preset scale and offset values for ADC-based current sensing.
  - Users should see more accurate current readings out of the box, reducing the need for manual calibration.
  - Improves telemetry reliability for power monitoring and consumption tracking on supported builds.
  - No changes required from users; existing setups will automatically use the new defaults when the ADC current meter source is selected.


kedeng:
> The scale and offset do not make sense. How were these values determined? Can you post the schematic for the current sensor on the included ESC? Do these values even need to be set in the target?

Sorry, I missed that recently.
These two values are based on that guide: https://betaflight.com/docs/development/Battery

Here are the related schematics:
ca504b82cc0674c219aae5539a0efcfb
2e9af5b3d5783bbb4592780f72ab8214

Yes. we need set it in target.

configs/SPEEDYBEEF405V5/config.h
```diff
@@ -119,6 +119,8 @@
 #define BOX_USER2_NAME "PROG PWR SWITCH"
 
 #define DEFAULT_PID_PROCESS_DENOM    2
+#define DEFAULT_CURRENT_METER_SCALE  27
+#define DEFAULT_CURRENT_METER_OFFSET -2644
```
ot0tot:
```suggestion

```

configs/SPEEDYBEEF405V5/config.h
```diff
@@ -119,6 +119,8 @@
 #define BOX_USER2_NAME "PROG PWR SWITCH"
 
 #define DEFAULT_PID_PROCESS_DENOM    2
+#define DEFAULT_CURRENT_METER_SCALE  27
```
ot0tot:
```suggestion
#define DEFAULT_CURRENT_METER_SCALE  360
```

ot0tot:
> > The scale and offset do not make sense. How were these values determined? Can you post the schematic for the current sensor on the included ESC? Do these values even need to be set in the target?
> 
> Sorry, I missed that recently. These two values are based on that guide: https://betaflight.com/docs/development/Battery
> 
> Here are the related schematics: ca504b82cc0674c219aae5539a0efcfb 2e9af5b3d5783bbb4592780f72ab8214
> 
> Yes. we need set it in target.

Review the data sheet for the current sensor again. Nowhere in your schematic are you generating an offset, and your scale does not match the values you have shown.

ot0tot:
The gain is 36, but that does not translate to a scale of 36. You are also calculating the offset incorrectly and with the wrong units.

ot0tot:
> @ot0tot This is the process by which we derived the scale and offset, as well as why we believe scale=36 is the theoretical value. Please take a look and see if there are any errors in our approach:
> 
> [ESC ammeter calibration.pdf](https://github.com/user-attachments/files/22070675/ESC.ammeter.calibration.pdf)

Set scale to 360 and offset to 0, and repeat your measurements.

Tetrahedron789:
Going to provide my input and experience with this ESC/FC. 
As of writing this, the default scale is 400 and offset 0. Readings were wildly off. A 4 minute flight showed 80mAh used with a max current of 5A. I ended up pulling about 1400mAh from the battery, based on how much was put back in after I charged it, and adjusted my scale to `35` with `new_scale = old scale * (OSD_mAh/actual_mAh)`, and readings seem to be more accurate as far as consumption goes. I did not test current accuracy with a multimeter

configs/SPEEDYBEEF405V5/config.h
kedeng: I have removed DEFAULT_CURRENT_METER_OFFSET

configs/SPEEDYBEEF405V5/config.h
kedeng: I changed this value to 36, because this is the theoretical value and it also matches the actual measured results

ot0tot (approved):

PR #880, opened 2025-08-24
Add optional MAG definitions for BETAFPVF405 and BETAFPVF405_ELRS targets

liyouluo:
Hi. Please check.

This configures the I2C interface on the board for use with an external magnetometer.

The target board has an I2C pad.

Thank you.

image

### DIFF
--- a/configs/BETAFPVF405/config.h
+++ b/configs/BETAFPVF405/config.h
@@ -39,6 +39,11 @@
 #define USE_FLASH
 #define USE_FLASH_W25Q128FV
 #define USE_MAX7456
+#ifndef USE_MAG
+#define USE_MAG
+#define USE_MAG_QMC5883
+#define USE_MAG_IST8310
+#endif
 
 #define BEEPER_PIN           PB4
 #define MOTOR1_PIN           PB0
@@ -60,6 +65,8 @@
 #define UART4_RX_PIN         PA1
 #define UART5_RX_PIN         PD2
 #define UART6_RX_PIN         PC7
+#define I2C1_SCL_PIN         PB8
+#define I2C1_SDA_PIN         PB9
 #define INVERTER_PIN_UART3   PC9
 #define INVERTER_PIN_UART6   PC8
 #define LED0_PIN             PB5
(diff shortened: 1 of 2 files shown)


configs/BETAFPVF405/config.h
```diff
@@ -39,6 +39,11 @@
 #define USE_FLASH
 #define USE_FLASH_W25Q128FV
 #define USE_MAX7456
+#ifndef USE_MAG
+#define USE_MAG
+#define USE_MAG_QMC5883
+#define USE_MAG_IST8310
+#endif
```
haslinghuis: If this board is not part of a BNF including MAG, this should be a cloud build option.
```suggestion
```

configs/BETAFPVF405_ELRS/config.h
```diff
@@ -39,6 +39,11 @@
 #define USE_FLASH
 #define USE_FLASH_W25Q128FV
 #define USE_MAX7456
+#ifndef USE_MAG
+#define USE_MAG
+#define USE_MAG_QMC5883
+#define USE_MAG_IST8310
+#endif
```
haslinghuis, line 46: Same as above

haslinghuis:
```suggestion
```

ot0tot:
PB8 is already mapped to ESC_SERIAL and RX_PPM in the target.  It cannot be used for those and I2C. 
Additionally, PB8 should be removed from the timer pin map in the BETAFPVF405_ELRS target.

haslinghuis:
@osirisinferi guess these targets were submitted before our manufacturer guideline requirements kicked in - as we all learn in the progress to get the best possible hardware out.

osirisinferi:
> @osirisinferi guess these targets were submitted before our manufacturer guideline requirements kicked in - as we all learn in the progress to get the best possible hardware out.

@haslinghuis Didn't mean to "point fingers", but thought maybe someone would like to fix these 🙂 Do you have schematics of the older boards?

haslinghuis:
No worrries, schematics were not required before the program.

PR #893, opened 2025-09-02
Add XROTOR-H7 (STM32H743) config – initial

Arwayday:
# Add target config for HW XROTOR-H7 (STM32H743).
Dual SPI ICM-42688P IMUs with external CLKIN; MPU6000 drivers kept for backward compatibility. Barometer on I2C2 (supports BMP280 / MS5611 / DPS310); external magnetometer can share I2C2. MAX7456 OSD on SPI3.
SDMMC (SDIO 4-bit) Blackbox enabled and set as default device. Motors on TIM1/TIM4, servos on TIM15, beeper PWM 3 kHz (inverted), LED strip enabled. Default ports: CRSF on USART1, MSP-DisplayPort on USART2, MSP on USART3, GPS on UART4, ESC telemetry on USART8. ADC: VBAT/CURR plus two external channels; defaults set (voltage scale 110, current scale 170, offset −350). Complete timer map and DMA options provided to avoid conflicts; DSHOT bitbang enabled. SWD enabled; PINIO defaults for “CAM 1,2” and “SW_VTX”.

## Summary by CodeRabbit

* **New Features**
  * Added support for HWH7 flight controller: DPS310 barometer and dual ICM-42688P gyros (clock-in capable).
  * Defaults RX to CRSF and provides many UARTs for RX, MSP, GPS, and ESC telemetry.
  * Supports up to 8 motors, 2 servos, LED/LED strip, beeper, and OSD.
  * SD card Blackbox logging (4-bit SDIO) and ADC-based voltage/current metering with presets.
  * Exposes user-configurable PINIO switches and enables DShot bitbang.


configs/XROTOR-H7/config.h
```diff
@@ -0,0 +1,207 @@
...
+- License along with this software.
+*
+- If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#pragma once
+
+#define FC_TARGET_MCU     STM32H743
+#define BOARD_NAME        XROTOR-H7
+#define MANUFACTURER_ID   HW
```
haslinghuis:
```suggestion
#define MANUFACTURER_ID   HOWI
```

configs/XROTOR-H7/config.h
```diff
@@ -0,0 +1,207 @@
...
+#define GYRO_2_EXTI_PIN               PC15
+#define GYRO_2_SPI_INSTANCE           SPI2
+#define GYRO_2_ALIGN                  CW270_DEG
+
+//UART
+#define UART1_TX_PIN                  PA9     //        
+#define UART1_RX_PIN                  PA10    //
+#define UART2_TX_PIN                  PA2     //
+#define UART2_RX_PIN                  PA3     //
+#define UART3_TX_PIN                  PD8     //BLUE TOOL
```
haslinghuis:
```suggestion
#define UART3_TX_PIN                  PD8     //BLUETOOTH
```

configs/XROTOR-H7/config.h
```diff
@@ -0,0 +1,207 @@
...
+// Debug
+#define USE_SWDIO_PIN                   PA13
+#define USE_SWCLK_PIN                   PA14
+
+//TIME MAP
+#define TIMER_PIN_MAPPING \
+    TIMER_PIN_MAP(0, MOTOR1_PIN, 1, 2) \
+    TIMER_PIN_MAP(1, MOTOR2_PIN, 1, 3) \
+    TIMER_PIN_MAP(2, MOTOR3_PIN, 1, 4) \
+    TIMER_PIN_MAP(3, MOTOR4_PIN, 1, 5) \
```
haslinghuis:
```suggestion
    TIMER_PIN_MAP(0, MOTOR1_PIN, 1, 1) \
    TIMER_PIN_MAP(1, MOTOR2_PIN, 1, 2) \
    TIMER_PIN_MAP(2, MOTOR3_PIN, 1, 3) \
    TIMER_PIN_MAP(3, MOTOR4_PIN, 1, 4) \
```

configs/XROTOR-H7/config.h
Arwayday: Done

configs/XROTOR-H7/config.h
Arwayday: Done

configs/XROTOR-H7/config.h
Arwayday: Done

configs/HWH7/config.h
```diff
@@ -0,0 +1,211 @@
...
+#define PINIO3_CONFIG 129
+
+//Default Module
+#define SERIALRX_PROVIDER        SERIALRX_CRSF
+#define SERIALRX_UART        SERIAL_PORT_USART1
+#ifdef USE_OSD_HD
+#define MSP_DISPLAYPORT_UART SERIAL_PORT_USART2
+#endif
+#define MSP_UART             SERIAL_PORT_USART3
+#define GPS_UART             SERIAL_PORT_UART4
```
haslinghuis:
```suggestion
#ifdef USE_GPS
#define GPS_UART             SERIAL_PORT_UART4
#endif
```

haslinghuis (approved):

configs/HWH7/config.h
```diff
@@ -0,0 +1,213 @@
...
+#define BOARD_NAME        HWH7
+#define MANUFACTURER_ID   HOWI
+
+//Gyroscope
+#define USE_ACC
+#define USE_ACC_SPI_MPU6000
+#define USE_ACC_SPI_ICM42688P
+
+#define USE_GYRO
+#define USE_GYRO_SPI_MPU6000
```
ot0tot: The schematic uses ICM42688P. You cannot use MPU6000 on the same footprint as they have different orientations. 
```suggestion
#define USE_ACC_SPI_ICM42688P

#define USE_GYRO
```

configs/HWH7/config.h
```diff
@@ -0,0 +1,213 @@
...
+#define USE_GYRO_SPI_ICM42688P
+
+//barometer
+#define USE_BARO
+#define USE_BARO_BMP280
+#define USE_BARO_MS5611
+#define USE_BARO_DPS310
```
ot0tot: The schematic only shows SPL06. Strongly suggest using a better baro. 
```suggestion
#define USE_BARO_DPS310
```

configs/HWH7/config.h
```diff
@@ -0,0 +1,213 @@
...
+    TIMER_PIN_MAP(4, MOTOR7_PIN, 1, -1) \
+    TIMER_PIN_MAP(5, GYRO_1_CLKIN_PIN, 1, -1) \
+    TIMER_PIN_MAP(6, MOTOR5_PIN, 1, -1) \
+    TIMER_PIN_MAP(7, MOTOR6_PIN, 1, -1) \
+    TIMER_PIN_MAP(8, GYRO_2_CLKIN_PIN, 1, -1) \
+    TIMER_PIN_MAP(9, MOTOR8_PIN, 1, -1) \
+    TIMER_PIN_MAP(10, SERVO1_PIN, 1, -1) \
+    TIMER_PIN_MAP(11, SERVO2_PIN, 1, -1) \
+    TIMER_PIN_MAP(12, BEEPER_PIN, 2, -1) \
+    TIMER_PIN_MAP(13, LED_STRIP_PIN, 2, 0)
```
ot0tot:
```suggestion
    TIMER_PIN_MAP( 0, MOTOR1_PIN, 1,  1) \
    TIMER_PIN_MAP( 1, MOTOR2_PIN, 1,  2) \
    TIMER_PIN_MAP( 2, MOTOR3_PIN, 1,  3) \
    TIMER_PIN_MAP( 3, MOTOR4_PIN, 1,  4) \
    TIMER_PIN_MAP( 4, MOTOR5_PIN, 1,  5) \
    TIMER_PIN_MAP( 5, MOTOR6_PIN, 1,  6) \
    TIMER_PIN_MAP( 6, MOTOR7_PIN, 1,  7) \
    TIMER_PIN_MAP( 7, MOTOR8_PIN, 1, -1) \
    TIMER_PIN_MAP( 8, SERVO1_PIN, 1, -1) \
    TIMER_PIN_MAP( 9, SERVO2_PIN, 1, -1) \
    TIMER_PIN_MAP(10, GYRO_1_CLKIN_PIN, 1, -1) \
    TIMER_PIN_MAP(11, GYRO_2_CLKIN_PIN, 1, -1) \
    TIMER_PIN_MAP(12, LED_STRIP_PIN, 2, 0)
```

configs/HWH7/config.h
```diff
@@ -0,0 +1,213 @@
...
+#define TIMUP4_DMA_OPT                7
+#define TIMUP5_DMA_OPT                8
+#define ADC1_DMA_OPT                  9
+#define ADC3_DMA_OPT                  10
+#define SPI1_TX_DMA_OPT               11
+#define SPI1_RX_DMA_OPT               12
+#define SPI2_TX_DMA_OPT               13
+#define SPI2_RX_DMA_OPT               14
+#define SPI3_TX_DMA_OPT               15
+#define SPI3_RX_DMA_OPT               16
```
ot0tot:
```suggestion
#define TIMUP1_DMA_OPT                8
#define TIMUP4_DMA_OPT                9
#define ADC1_DMA_OPT                  10
#define ADC3_DMA_OPT                  11
```

configs/HWH7/config.h
```diff
@@ -0,0 +1,213 @@
...
+#define SERIALRX_PROVIDER        SERIALRX_CRSF
+#define SERIALRX_UART        SERIAL_PORT_USART1
+#ifdef USE_OSD_HD
+#define MSP_DISPLAYPORT_UART SERIAL_PORT_USART2
+#endif
+#define MSP_UART             SERIAL_PORT_USART3
+#ifdef USE_GPS
+#define GPS_UART             SERIAL_PORT_UART4
+#endif
+#define ESC_SENSOR_UART      SERIAL_PORT_USART8
```
ot0tot: UART8 is broken out to pads.
```suggestion

```

configs/HWH7/config.h
```diff
@@ -0,0 +1,213 @@
...
+
+//ADC
+#define ADC_VBAT_PIN                  PC0        //ADC123_INP10,
+#define ADC_CURR_PIN                  PC1        //ADC123_INP11,
+#define ADC_EXTERNAL1_PIN             PC4        //ADC12_INP4,   
+#define ADC_EXTERNAL2_PIN             PC5        //ADC12_INP8,   CU2
+
+#define DEFAULT_CURRENT_METER_SOURCE CURRENT_METER_ADC
+#define DEFAULT_VOLTAGE_METER_SOURCE VOLTAGE_METER_ADC
+#define DEFAULT_VOLTAGE_METER_SCALE_DEFAULT 110
```
ot0tot: This is the default
```suggestion

```

configs/HWH7/config.h
```diff
@@ -0,0 +1,213 @@
...
+#define DEFAULT_CURRENT_METER_SCALE         170
+#define DEFAULT_CURRENT_METER_OFFSET       -350
```
ot0tot: Not an AIO, these depend on the ESC.
```suggestion

```

configs/HWH7/config.h
```diff
@@ -0,0 +1,213 @@
...
+#define BARO_I2C_INSTANCE             I2CDEV_2
+#define MAG_I2C_INSTANCE              I2CDEV_2
+
+//LED
+#define LED0_PIN                      PB2 
+#define LED_STRIP_PIN                 PA0       //TIM5_CH1
+
+//BEEPER
+#define BEEPER_PIN                    PA1       //TIM5_CH2
+#define BEEPER_PWM_HZ                 3000
```
ot0tot, line 47: Passive buzzer not supported because timer conflicts with LED_STRIP
```suggestion

```

configs/HWH7/config.h
Arwayday: No problem, these barometers are outdated. I originally included them for maximum compatibility, but it seems they are no longer necessary.

configs/HWH7/config.h
Arwayday: Friend, I don’t understand why this modification is necessary. I believe motor 8 cannot use DMA, so even if DMA is configured for motors 5/6/7, it doesn’t really provide any advantage. My original configuration was generated directly from the timer map directive, and I’ve already verified that it works without issues.

configs/HWH7/config.h
Arwayday: Friend, I cannot make this modification. First, removing TIMUP5_DMA_OPT will cause the LED_STRIP to stop working. Second, enabling DMA for SPI peripherals can indeed improve performance — but could you please explain why it cannot be used this way?

configs/HWH7/config.h
Arwayday: Although UART8 is brought out to pads, this is provided only as a safeguard against damage to the ESC connector. By default it is wired to the ESC port and configured to receive telemetry data.

configs/HWH7/config.h
Arwayday: I understand this is the default voltage divider ratio. The voltage divider on my schematic is also set to 110, so I don't see any issue.

configs/HWH7/config.h
Arwayday: Friend, I understand that the current scaling and offset depend on the ESC, but flight controllers are typically sold together with our ESCs. I’d like to set a default current-scale/offset value on the FC; this would be convenient for users who purchase our bundled kits.

ot0tot: There should be no offset unless you are using a Hall effect sensor. Can you provide the schematic for the ESC current sensor?

Arwayday: Apologies — you’re right, this is indeed an issue. I’ve just consulted with my colleagues and will submit a follow-up update to fix it. There should be no offset.

configs/HWH7/config.h
Arwayday: Resolved

configs/HWH7/config.h
Arwayday: I’ve pushed a follow-up commit that fixes the erroneous current offset and scale values.
#define DEFAULT_CURRENT_METER_SCALE         117
#define DEFAULT_CURRENT_METER_OFFSET        0

configs/HWH7/config.h
Arwayday: Timer 5 supports four capture/compare channels. I have assigned channel 2 to the beeper and channel 1 to the LED strip. So far everything is running normally with no issues.

osirisinferi: Multiple channels of the same timer would run on the same frequency. And WS2812B runs on a completely different frequency than 3000 Hz, 800 kHz to be exact. Thus, it wouldn't/shouldn't be possible to run a WS2812B LED strip and a passive buzzer at the same time.

Isn't TIM2 an option for either of those PA0/PA1 pins? TIM2 seems to be a general-purpose timers comparable to TIM5 and I don't see any pin associated to TIM2 already according to the inline comments. (Haven't double checked.)

configs/HWH7/config.h
ot0tot: It does not need to be included in the config because it is already the default value.

configs/HWH7/config.h
ot0tot: TIMUP has nothing to do with LED_STRIP.
DMA for SPI is automatic and does not need to be defined here

configs/HWH7/config.h
ot0tot: ESC telemetry via UART has been superseded by DShot EDT. If someone really wants to use this, they can set it in the ports tab.

configs/HWH7/config.h
ot0tot: This is necessary because your timer map is wrong.

configs/HWH7/config.h
Arwayday: Understood — that explanation makes sense to me. Thank you for your guidance.

configs/HWH7/config.h
Arwayday: That makes sense — the duplicate definition here is redundant. Thank you for the guidance.

configs/HWH7/config.h
Arwayday: Hi there — I’ve updated the config: removed incorrect macros and redundant settings. When you have a moment, I’d appreciate a quick review.

configs/HWH7/config.h
Arwayday: Thank you for pointing this out. I’ve pushed a new commit—when you have a moment, could you please take another look?

configs/HWH7/config.h
Arwayday: Mystery solved. The beeper gave one startup “peep,” and then LED_STRIP grabbed the timer like a seagull stealing fries—no more beeps after that.
Big shout-out to osirisinferi for spotting TIM2. I’ve moved LED_STRIP there, and peace has been restored: beeper beeps, LEDs blink, no more turf wars.

configs/HWH7/config.h
```diff
@@ -0,0 +1,201 @@
...
+
+//ICM Clock
+#define USE_GYRO_CLKIN
+#define GYRO_1_CLKIN_PIN              PB8     // TIM16_CH1,
+#define GYRO_2_CLKIN_PIN              PB9     // TIM17_CH1
+
+#define DEFAULT_GYRO_TO_USE          GYRO_CONFIG_USE_GYRO_BOTH
+
+//ICM-42688 (1)
+#define ENSURE_MPU_DATA_READY_IS_LOW          // 防止重复中断
```
osirisinferi: Since [932d83a](https://github.com/betaflight/config/pull/893/commits/932d83a5f316afece665c473c58d6ef9eb7226d0) I don't think this define is necessary any longer.
```suggestion
```

Arwayday: No problem. This macro definition was only set because I saw many people adding it this way; if it's not necessary, it can be deleted.

configs/HWH7/config.h
```diff
@@ -0,0 +1,201 @@
...
+#define ADC_EXTERNAL2_PIN             PC5        // ADC12_INP8,   CU2
+
+#define DEFAULT_CURRENT_METER_SOURCE CURRENT_METER_ADC
+#define DEFAULT_VOLTAGE_METER_SOURCE VOLTAGE_METER_ADC
+#define DEFAULT_CURRENT_METER_SCALE         117
+#define DEFAULT_CURRENT_METER_OFFSET        0
+
+// Debug
+#define USE_SWDIO_PIN                   PA13
+#define USE_SWCLK_PIN                   PA14
```
osirisinferi, line 148: I can't find any reference to `USE_SWDIO_PIN` nor `USE_SWCLK_PIN` in the BetaFlight code base, are these really necessary/doing anything? Perhaps for a modified/custom use BF FW?

Arwayday: This is a macro definition reserved for the 4.6.0 firmware.

Arwayday: I consulted V22 about this matter before; using the latest master version to compile will allow the use of this macro definition.

Arwayday: These two macros may be completely useless, because the user has no idea how to use them. I just added them since I saw others using them.

osirisinferi: @Arwayday I've learned such defines are not actually doing anything in the firmware, but are there to indicate the fact those pins are broken out on the board (e.g., available using test pads).

If you don't have any SWDIO/SWCLK pads connecting to PA13/PA14 on your board, you should remove them, otherwise it's fine.

configs/HWH7/config.h
ot0tot: Need to remove this define

configs/HWH7/config.h
```diff
@@ -0,0 +1,200 @@
...
+    TIMER_PIN_MAP(11, GYRO_2_CLKIN_PIN, 1, -1) \
+    TIMER_PIN_MAP(12, BEEPER_PIN, 2, -1) \
+    TIMER_PIN_MAP(13, LED_STRIP_PIN, 1, 0)
+
+#define DEFAULT_DSHOT_BITBANG           DSHOT_BITBANG_ON
+
+//DMA
+#define TIMUP1_DMA_OPT                8
+#define TIMUP4_DMA_OPT                9
+#define TIMUP5_DMA_OPT                10
```
ot0tot:
```suggestion
```

configs/HWH7/config.h
```diff
@@ -0,0 +1,200 @@
...
+#define PINIO2_CONFIG 129
+#define BOX_USER2_NAME       "SW_VTX"
+
+#define PINIO3_PIN PD10
+#define PINIO3_BOX 0
+#define PINIO3_CONFIG 129
+
+//Default Module
+#define SERIALRX_PROVIDER        SERIALRX_CRSF
+#define SERIALRX_UART            SERIAL_PORT_USART1
```
ot0tot: UART1 is susceptible to DFU hijacking and cannot be used for USB powered peripherals, e.g. receiver or GPS
```suggestion

```

Arwayday: No problem—this could indeed cause confusion for users when they try to use DFU mode.

configs/HWH7/config.h
Arwayday: We currently have connected test pads in place to facilitate programming during production. Thank you for your patient guidance.

configs/HWH7/config.h
Arwayday: I’ve remapped LED_STRIP to TIM2 and BEEPER to TIM5. With this mapping, the current macro looks valid on my side. I built and tested locally and everything functions as expected. Do you still recommend any changes?   @ot0tot @haslinghuis

haslinghuis: Should be sorted.

osirisinferi (commented):
LGTM :+1:

Arwayday:
> LGTM 👍

configs/HWH7/config.h
```diff
@@ -0,0 +1,198 @@
...
+//ADC
+#define ADC_VBAT_PIN                  PC0        // ADC123_INP10,
+#define ADC_CURR_PIN                  PC1        // ADC123_INP11,
+#define ADC_EXTERNAL1_PIN             PC4        // ADC12_INP4,   
+#define ADC_EXTERNAL2_PIN             PC5        // ADC12_INP8,   CU2
+
+#define DEFAULT_CURRENT_METER_SOURCE CURRENT_METER_ADC
+#define DEFAULT_VOLTAGE_METER_SOURCE VOLTAGE_METER_ADC
+#define DEFAULT_CURRENT_METER_SCALE         117
+#define DEFAULT_CURRENT_METER_OFFSET        0
```
ot0tot: This is already the default and does not need to be included. 
```suggestion
```

Arwayday: Duplicate macros have been removed. Thank you for your careful and patient guidance. When you have a moment, could you please review and re-approve this PR?

PR #917, opened 2025-09-26
Add the missing UART3

YI-BOYANG:
Upon inspection, it was found that the definition of UART3 was missing.

## Summary by CodeRabbit

* **New Features**
  * Enabled UART3 on the GEPRCF722 BT HD V3 target by adding TX/RX pin mapping, providing an additional serial port for peripherals (e.g., receivers, telemetry, GPS).
  * Improves flexibility and peripheral compatibility without altering existing UART configurations or other board settings.


haslinghuis (approved):

ot0tot:
The UART is not missing, you are using those pins for I2C and they are connected to the barometer.

osirisinferi:
@coderabbitai Why did you not detect the duplicate pin assignments which would result from this PR?

PR #918, opened 2025-09-29
Update TAKERF722SE

YI-BOYANG:
Added bmp280 and dps310, set the default UART, and added the configuration for pinio1.

## Summary by CodeRabbit

- New Features
  - Added barometer support (BMP280, DPS310) on TAKERF722SE.
  - Enabled MSP DisplayPort over UART1 for on-screen display integrations.

- Changes
  - Default Serial RX moved to UART2 (was UART3).
  - Removed BMI270 accelerometer/gyroscope configuration in favor of barometer-focused options.


configs/TAKERF722SE/config.h
```diff
@@ -103,9 +105,10 @@
 #define ADC1_DMA_OPT 0
 
 #define DEFAULT_FEATURES FEATURE_ESC_SENSOR
-#define SERIALRX_UART SERIAL_PORT_USART3
 
-// serial 5 1024 115200 57600 0 115200
+#define MSP_DISPLAYPORT_UART            SERIAL_PORT_USART1
+#define SERIALRX_UART                   SERIAL_PORT_USART2
```
VitroidFPV, line 110: Default UART port functions are generally only used for peripherals that are integrated onto the board itself, or BNF quads where the factory setup will consistently use those specific ports. If you cannot ensure consistency, it's best to leave this up to the user or factory setup

YI-BOYANG: imageIn the design of FC, UART2 is specifically used for RX communication. However, the target has not been modified before.

VitroidFPV: If that's how it comes set up in the BNFs that use it, that's probably fine

configs/TAKERF722SE/config.h
```diff
@@ -116,7 +119,7 @@
 #define BEEPER_INVERTED
 #define MAX7456_SPI_INSTANCE SPI2
 #define PINIO1_CONFIG 129
-#define PINIO1_BOX 0
+#define PINIO1_BOX 40
```
ot0tot: Why are you changing this? What does PINIO 1 control?

YI-BOYANG: Because some users have requested that USER1 be used to control the on/off of the 12V power supply.

ot0tot: What is PINIO1 connected to on the FC? Can you post the schematic in the Discord?

YI-BOYANG: QQ_1759201293471 This is the schematic diagram of the PINIO1 connection. It can control a pad to output 12V voltage.

ot0tot: So for the past several years the VTX power on this FC was tied to the arm switch? That makes no sense. Have you changed the design?

YI-BOYANG: In fact, the previous VTX power supply and ARM switch were not related. I don't understand why pnio1_box was set to 0 when designing the target. Based on the user's feedback, we are now considering adjusting the design and using PIO1 to control the 12V power supply.

ot0tot: What was pin C01 connected to previously?

YI-BOYANG: Prior to this, PC1 was not connected to any location. Therefore, this change will not have any impact on the existing FC design.

ot0tot: If you are making changes to the FC design, this requires a new target.

YI-BOYANG: I believe that the new design and the old design can use the same target; there is no need to create a new one. However, the new design is still under planning. If you don't agree with the modifications made by pinio1, I can first cancel them and then we can have a discussion later. We need to add a barometer now.

ot0tot: Does this FC have a barometer? The product page and images make no mention of it.
https://geprc.com/product/geprc-taker-f722-se-flight-controller/

image

YI-BOYANG: Our subsequent plan is to add the "spl06-001 baro" to it.

ot0tot: 1) The SPL06 is a poor quality baro, use something better.
2) Any revisions to the FC require a new review.  The design as-is does not meet the current standards.

PR #931, opened 2025-10-22
Fix HUMMINGBIRD_FC305 current meter scale calibration

LYNHQQ:
This change improves the current meter scale to provide more accurate current measurements for the HUMMINGBIRD_FC305 flight controller. The previous value of 800 was causing less accurate current readings, and the new value of 310 has been tested and verified to provide relatively more accurate measurements.

## Summary by CodeRabbit

* **Chores**
  * Adjusted current meter scaling for the HUMMINGBIRD_FC305 hardware variant to change how current measurements are scaled, improving reported current values and alignment with expected readings. No other behaviors or features were modified.


LYNHQQ:
> According to the schematic, this FC does not have an ESC. The current sensor scale should not be included in the config.

It's an integral part of the ammeter's ADC, and its value corresponds directly to the ammeter's scaling factor. In my view, it should be retained in the configuration. Otherwise, users will have to perform the calibration themselves, which would be inconvenient and complicated.

ot0tot:
But the FC can be used with any ESC. Manufacturers should include the current scale value with their ESC documentation.

LYNHQQ:
> But the FC can be used with any ESC. Manufacturers should include the current scale value with their ESC documentation.

Most customers who purchase this flight controller get the stack, which is paired with our ESC. If our firmware can include this setting, it would greatly reduce the hassle for customers. Even if customers use other ESCs, they would still need to make adjustments, but at least the customers using our stack can use it directly. Overall, this is for the customer's convenience and will not cause any negative impact

ot0tot:
I did a quick search and only saw the FC alone. I see now that there is also a stack available, and that your documentation is missing the ESC current sensor scale.

How was the value off by so much? Can you post the current sensor schematic from the ESC?

LYNHQQ:
> I did a quick search and only saw the FC alone. I see now that there is also a stack available, and that your documentation is missing the ESC current sensor scale.
> 
> How was the value off by so much? Can you post the current sensor schematic from the ESC?

Hi @ot0tot , please check our schematic. In my actual testing, I found that 310 is a relatively accurate current meter scale. If you have a better suggestion, I am also very willing to re-test.
image

ot0tot:
Based on your schematic, the scale would be 250.

PR #961, opened 2025-11-18
add foxeerf405v2 default flash

johannes-do:
Otherwise the blackblox flash is not being picked up and shows: "No dataflash chip found".

## Summary by CodeRabbit

* **Chores**
  * Updated the default configuration for the FOXEERF405V2 flight controller to use onboard flash for blackbox recording, ensuring flight logs are stored to flash by default.


configs/FOXEERF405V2/config.h
```diff
@@ -101,6 +101,7 @@
 
 #define ADC1_DMA_OPT 1
 
+#define DEFAULT_BLACKBOX_DEVICE BLACKBOX_DEVICE_FLASH
```
haslinghuis: Perhaps move to around line 111-112

haslinghuis (approved):

configs/FOXEERF405V2/config.h
johannes-do: Yes, makes sense.

PR #1074, opened 2026-04-09
Add HWF435 target

Hebo98:
## Summary by CodeRabbit

* **New Features**
  * Added support for the HWF435 flight controller board.
  * Enables onboard sensors and peripherals: SPI IMU (selectable accel/gyro families), DPS310 barometer, magnetometer, external flash storage, and OSD.
  * Configures motor/servo outputs, LED/strip, beeper, multiple UART/I2C/SPI ports, battery/current ADC, and sensible defaults for Blackbox, mixer, serial RX, voltage/current scaling, and DShot.

[Review Change Stack](https://app.coderabbit.ai/change-stack/betaflight/config/pull/1074)


configs/HWF435/config.h
```diff
@@ -0,0 +1,151 @@
...
+
+#define FC_TARGET_MCU                AT32F435G
+#define BOARD_NAME                   HWF435
+#define MANUFACTURER_ID              HOWI
+#define SYSTEM_HSE_MHZ               8
+
+#define USE_GYRO
+#define USE_ACC
+#define USE_GYRO_SPI_MPU6500
+#define USE_ACC_SPI_MPU6500
```
haslinghuis: Please exclude MPU6500

Hebo98:
```suggestion
```

configs/HWF435/config.h
```diff
@@ -0,0 +1,149 @@
...
+#define UART3_RX_PIN                 PB11
+#define UART4_TX_PIN                 PH3
+#define UART4_RX_PIN                 PA1
+#define UART7_TX_PIN                 PC0
+#define UART7_RX_PIN                 PC1
+#define UART8_TX_PIN                 PC2
+#define UART8_RX_PIN                 PC3
+
+#define USE_MAG
+#define USE_GPS
```
osirisinferi: These are user defines and should not be included in the `config.h` unless there are actual magnetos or GNSS modules on the FC. This is not the case according to the schematic.

```suggestion
```

configs/HWF435/config.h
```diff
@@ -0,0 +1,149 @@
...
+    TIMER_PIN_MAP( 3, PC9, 2, 3)  \
+    TIMER_PIN_MAP( 4, PB0, 2, 4)  \
+    TIMER_PIN_MAP( 5, PB1, 2, 5)  \
+    TIMER_PIN_MAP( 6, PB4, 1, 6)  \
+    TIMER_PIN_MAP( 7, PB5, 1, 7)  \
+    TIMER_PIN_MAP( 8, PB6, 1, -1) \
+    TIMER_PIN_MAP( 9, PB7, 1, -1) \
+    TIMER_PIN_MAP(10, PB3, 1, -1) \
+    TIMER_PIN_MAP(11, PH2, 1, 11) \
+    TIMER_PIN_MAP(12, PA8, 1, -1)
```
osirisinferi: Let's add the beeper pin on TMR20_CH1. This should work. If it gives issues at testing the hardware samples for some reason we either need to fix some bug or we can remove it again afterwards.

```suggestion
    TIMER_PIN_MAP(12, PA8, 1, -1) \
    TIMER_PIN_MAP(13, PB2, 2, -1)
```

configs/HWF435/config.h
```diff
@@ -0,0 +1,149 @@
...
+
+#define DEFAULT_BLACKBOX_DEVICE        BLACKBOX_DEVICE_FLASH
+
+#define DEFAULT_VOLTAGE_METER_SOURCE    VOLTAGE_METER_ADC
+#define DEFAULT_CURRENT_METER_SOURCE    CURRENT_METER_ADC
+#define DEFAULT_VOLTAGE_METER_SCALE     168
+#define DEFAULT_CURRENT_METER_SCALE     117
+
+#define DEFAULT_MIXER                   MIXER_OCTOX8
+#define DEFAULT_DSHOT_BITBANG           DSHOT_BITBANG_OFF
```
osirisinferi: Why is bitbanged DShot off?
```suggestion
#define DEFAULT_DSHOT_BITBANG           DSHOT_BITBANG_ON
```

configs/HWF435/config.h
```diff
@@ -0,0 +1,149 @@
...
+
+
+#define DEFAULT_BLACKBOX_DEVICE        BLACKBOX_DEVICE_FLASH
+
+#define DEFAULT_VOLTAGE_METER_SOURCE    VOLTAGE_METER_ADC
+#define DEFAULT_CURRENT_METER_SOURCE    CURRENT_METER_ADC
+#define DEFAULT_VOLTAGE_METER_SCALE     168
+#define DEFAULT_CURRENT_METER_SCALE     117
+
+#define DEFAULT_MIXER                   MIXER_OCTOX8
```
osirisinferi, line 140: Is this board going to be shipped in a BNF octocopter? Or otherwise specifically meant for octocopters?

configs/HWF435/config.h
```diff
@@ -0,0 +1,149 @@
...
+    TIMER_PIN_MAP(12, PA8, 1, -1)
+
+
+
+#define DEFAULT_BLACKBOX_DEVICE        BLACKBOX_DEVICE_FLASH
+
+#define DEFAULT_VOLTAGE_METER_SOURCE    VOLTAGE_METER_ADC
+#define DEFAULT_CURRENT_METER_SOURCE    CURRENT_METER_ADC
+#define DEFAULT_VOLTAGE_METER_SCALE     168
+#define DEFAULT_CURRENT_METER_SCALE     117
```
osirisinferi, line 138: Is this FC going to be part of a specific stack with a specific ESC?

configs/HWF435/config.h
```diff
@@ -0,0 +1,149 @@
...
+
+#define DEFAULT_MIXER                   MIXER_OCTOX8
+#define DEFAULT_DSHOT_BITBANG           DSHOT_BITBANG_OFF
+#define DEFAULT_MOTOR_DSHOT_SPEED       PWM_TYPE_DSHOT600
```
osirisinferi: This MCU is not powerfull enough to run at a PID loop of 8 kHz and when running at 4 kHz, DShot600 does not have any benefit above DShot300.

```suggestion
#define DEFAULT_MOTOR_DSHOT_SPEED       PWM_TYPE_DSHOT300
#define DEFAULT_PID_PROCESS_DENOM        2
```

configs/HWF435/config.h
```diff
@@ -0,0 +1,149 @@
...
+
+#define DEFAULT_ALIGN_BOARD_ROLL        180
+#define DEFAULT_ALIGN_BOARD_PITCH       180
+#define DEFAULT_ALIGN_BOARD_YAW         270
```
osirisinferi: Isn't this the same as `DEFAULT_ALIGN_BOARD_YAW` = 90?

I'm also curious why the board would need to be rotated 90° clockwise.

configs/HWF435/config.h
```diff
@@ -0,0 +1,149 @@
...
+
+#define DEFAULT_RX_FEATURE              FEATURE_RX_SERIAL
+#define SERIALRX_PROVIDER               SERIALRX_CRSF
+#define SERIALRX_UART                   SERIAL_PORT_USART7
+#define OSD_DISPLAYPORT_DEVICE          MAX7456
```
osirisinferi: Not necessary, will automatically be used if the firmware is built with the OSD (SD) define.
```suggestion
```

configs/HWF435/config.h
Hebo98: Yes, this flight controller is designed to be used together with a specific ESC.

configs/HWF435/config.h
Hebo98: I verified it through testing, and it is indeed equivalent. I’ve made the changes. Thank you for your guidance. Also, the arrow direction on the PCB does need to be rotated by this angle.

configs/HWF435/config.h
Hebo98: Yes，We have related usage needs.

PR #1098, opened 2026-05-07
Add Nexus XR Wing target

ot0tot:
Add Radiomaster Nexus XR Wing target

@coderabbitai note this is a Wing FC, and the usual Betaflight quad FC restrictions do not apply. 

- [ ] Hardware test

## Summary by CodeRabbit

* **New Features**
  * Added support for the NEXUSXR flight controller board: motor and optional servo/wing control, multiple UART ports (including built‑in RC telemetry), onboard IMU and barometer, onboard SPI flash for blackbox, and default telemetry/blackbox settings.
  * Enabled ADC-based current/voltage sensing with sensible scaling, DShot ESC protocol support, and default sensor/telemetry routing for out-of-the-box use.

### DIFF
--- a/configs/NEXUSXR/config.h
+++ b/configs/NEXUSXR/config.h
@@ -49,6 +49,10 @@
 #define SERVO1_PIN           PB4  // S1
 #define SERVO2_PIN           PB5  // S2
 #define SERVO3_PIN           PB0  // S3
+#define SERVO4_PIN           PA2  // RPM
+#define SERVO5_PIN           PA3  // TLM
+#define SERVO6_PIN           PB6  // AUX
+#define SERVO7_PIN           PB7  // SBUX
 
 #define UART1_TX_PIN         PB6 // AUX or S6
 #define UART1_RX_PIN         PB7 // SBUS or S7
--- a/configs/NEXUSXR/config.h
+++ b/configs/NEXUSXR/config.h
@@ -1,0 +1,126 @@
+/*
+ * This file is part of Betaflight.
+ *
+ * Betaflight is free software. You can redistribute this software
+ * and/or modify this software under the terms of the GNU General
+ * Public License as published by the Free Software Foundation,
+ * either version 3 of the License, or (at your option) any later
+ * version.
+ *
+ * Betaflight is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ *
+ * See the GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public
+ * License along with this software.
+ *
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#define FC_TARGET_MCU     STM32F7X2
+
+#define BOARD_NAME        NEXUSXR
+#define MANUFACTURER_ID   RDMS
+
+#define USE_ACC
+#define USE_GYRO
+#define USE_GYRO_CLKIN
+#define USE_GYRO_SPI_ICM42688P
+#define USE_ACC_SPI_ICM42688P
+#define USE_BARO
+#define USE_BARO_DPS310
+#define USE_FLASH
+#define USE_FLASH_W25N
+
+#ifndef USE_WING
+#define USE_WING
+#endif
+
+#ifndef USE_SERVOS
+#define USE_SERVOS
+#endif
+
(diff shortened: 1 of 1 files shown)


ot0tot:
@coderabbitai review and approve

haslinghuis:
@coderabbitai approve

haslinghuis (approved):

configs/NEXUSXR/config.h
```diff
@@ -0,0 +1,131 @@
...
+#define USE_FLASH
+#define USE_FLASH_W25N
+
+#ifndef USE_WING
+#define USE_WING
+#endif
+
+#ifndef USE_SERVOS
+#define USE_SERVOS
+#endif
```
haslinghuis, line 45: USE_WING should already include USE_SERVOS - but being explicit does not hurt.

configs/NEXUSXR/config.h
```diff
@@ -0,0 +1,131 @@
...
+    TIMER_PIN_MAP( 6, PA3,  2, -1 ) /* TIM5_CH4 */ \
+    TIMER_PIN_MAP( 7, PB6,  1, -1 ) /* TIM4_CH1 */ \
+    TIMER_PIN_MAP( 8, PB7,  1, -1 ) /* TIM4_CH2 */ \
+    TIMER_PIN_MAP( 9, PB9,  2, -1 ) /* TIM11_CH1 */ 
+
+#define ADC1_DMA_OPT                     0
+
+#define BARO_I2C_INSTANCE                I2CDEV_3
+#define MAG_I2C_INSTANCE                 I2CDEV_3
+#define ADC_INSTANCE                     ADC1
```
haslinghuis, line 119: ADC1 is default - but again - nothing wrong to be explicit

configs/NEXUSXR/config.h
```diff
@@ -0,0 +1,131 @@
...
+#define DEFAULT_BLACKBOX_DEVICE          BLACKBOX_DEVICE_FLASH
+#define DEFAULT_DSHOT_BITBANG            DSHOT_BITBANG_ON
+#define DEFAULT_CURRENT_METER_SOURCE     CURRENT_METER_ADC
+#define DEFAULT_VOLTAGE_METER_SOURCE     VOLTAGE_METER_ADC
+#define DEFAULT_VOLTAGE_METER_SCALE      238
+#define DEFAULT_VOLTAGE_METER_MULTIPLIER 2
+#define DEFAULT_CURRENT_METER_SCALE      273
+#define PINIO1_BOX                       40
+#define PINIO1_CONFIG                    129
+#define BOX_USER1_NAME                   "ELRS OFF"
```
haslinghuis, line 129: This would be good practice for boards to avoid DFU hijack ?

ot0tot: No, the better solution is to just not use those pins for receiver/GPS. 

I think this would add an unnecessary level of complexity and lead to support headaches. It would also only work with built-in receivers, and require a spare GPIO.

configs/NEXUSXR/config.h
ot0tot: Yes, just following the template we used for other Wing FCs. I'm fine with removing though, as it is redundant.

ot0tot:
Do not merge until we have confirmation from RadioMaster that it is working properly with the hardware. We currently do not have hardware to test.

configs/NEXUSXR/config.h
```diff
@@ -0,0 +1,131 @@
...
+    TIMER_PIN_MAP( 5, PA2,  2, -1 ) /* TIM5_CH3 */ \
+    TIMER_PIN_MAP( 6, PA3,  2, -1 ) /* TIM5_CH4 */ \
+    TIMER_PIN_MAP( 7, PB6,  1, -1 ) /* TIM4_CH1 */ \
+    TIMER_PIN_MAP( 8, PB7,  1, -1 ) /* TIM4_CH2 */ \
+    TIMER_PIN_MAP( 9, PB9,  2, -1 ) /* TIM11_CH1 */ 
+
+#define ADC1_DMA_OPT                     0
+
+#define BARO_I2C_INSTANCE                I2CDEV_3
+#define MAG_I2C_INSTANCE                 I2CDEV_3
```
ot0tot: I2C pins are not broken out. This should probably be moved to I2C2, which will share pins with UART3 (Port C).

sugaarK (commented):
tested on hardware sent from RM

sugaarK (approved):

configs/NEXUSXR/config.h
ot0tot: Resolved.

PR #1118, opened 2026-06-04
Fix SPEEDYBEEF405AIO DEFAULT_ALIGN_BOARD_YAW

gwlim:
The default align board yaw in SPEEDYBEEF405AIO is incorrect. I have the Bee25 and encountered the same issue as the issue below

Fixes the issue below
https://github.com/betaflight/betaflight-configurator/issues/4924

Screenshot From 2026-06-04 23-07-12
Screenshot From 2026-06-04 23-07-26

## Summary by CodeRabbit

* **Chores**
  * Updated board orientation alignment configuration for improved device calibration.


gwlim:
@haslinghuis please merge this otherwise owners of bee25 who upgrade their firmware will think upgrading their firmware broke their drone

haslinghuis (approved):

nerdCopter (approved):

osirisinferi:
Is the `SPEEDYBEEF405AIO` only used for this Bee25?

Edit: I'm also wondering why upgrading to 2025.12 might have influenced this, as mentioned in the issue you've opened regarding this 🤔

nerdCopter:
> Is the `SPEEDYBEEF405AIO` only used for this Bee25?

good question, are they simply replacing ICM42688p with BMI270? if that's the case then broad alignment redefine is not the solution.

osirisinferi:
> if that's the case then broad alignment redefine is not the solution.

Indeed. Need more info regarding that.

I'm also wondering if *board* alignment even should have a place in the target config file, as we should assume the target is used accordingly to the arrow on the board itself. The target should just have the correct non-user configurable gyro alignment set accordingly to the arrow. Board alignment should a user setting. Unless there is some convention that board alignments are put into targets if a pre-build drone requires this.

Edit: board doesn't have a BMI270 defined 🤔 We should first know if this indeed is a change between 4.5.x and 2025.12.x and where and how before changing anything.

haslinghuis:
See https://github.com/betaflight/betaflight-configurator/issues/4924#issuecomment-4124351427

PR #1119, opened 2026-06-04
Add UPT1 rangefinder and optical flow

SteveCEvans:
Explicit inclusion of rangefinder and optical flow now required.

## Release Notes

* **New Features**
  * Enabled optical flow functionality for motion detection
  * Enabled rangefinder functionality for distance measurement

### DIFF
--- a/configs/CADDX_PROTOS_F4/config.h
+++ b/configs/CADDX_PROTOS_F4/config.h
@@ -123,3 +123,8 @@
 #define ADC1_DMA_OPT        1
 #define ADC_INSTANCE ADC1
 
+//--------OPTICAL FLOW-------
+#define USE_RANGEFINDER
+#define USE_RANGEFINDER_UPT1
+#define USE_OPTICALFLOW
+#define USE_OPTICALFLOW_UPT1


haslinghuis (approved):
Does it not conflict when also use the cloud build options ?

sugaarK (approved):

osirisinferi:
> Does it not conflict when also use the cloud build options ?

Probably. When I build locally with `-DCLOUD_BUILD -DUSE_RANGEFINDER -DUSE_OPTICALFLOW` it errors out as expected:

```
In file included from ./src/main/platform.h:26,
                 from ./src/platform/STM32/rcc_stm32.c:21:
./src/config/configs/CADDX_PROTOS_F4/config.h:127: error: "USE_RANGEFINDER" redefined [-Werror]
  127 | #define USE_RANGEFINDER
      | 
<command-line>: note: this is the location of the previous definition
./src/config/configs/CADDX_PROTOS_F4/config.h:129: error: "USE_OPTICALFLOW" redefined [-Werror]
  129 | #define USE_OPTICALFLOW
      | 
<command-line>: note: this is the location of the previous definition
cc1: all warnings being treated as errors
```

Just enabling RANGEFINDER (with a guard) should be enough, as `common_post.h` enables the UPT1 driver already:

```
#if defined(USE_RANGEFINDER)

(...)
#ifndef USE_RANGEFINDER_UPT1
#define USE_RANGEFINDER_UPT1
#endif

#endif // USE_RANGEFINDER
```

But it's properly guarded, so it doesn't hurt to have it included explicitely.

PR #1141, opened 2026-07-05
fix: remove non-existent USE_BARO_LPS define from four targets

haslinghuis:
## Summary

Removes the dead `USE_BARO_LPS` define from four targets. `USE_BARO_LPS` is **not referenced anywhere in the firmware** — only `USE_BARO_SPI_LPS` (LPS25, SPI) and `USE_BARO_LPS22DF` (LPS22, I2C) exist. So on these boards the bare `USE_BARO_LPS` is a no-op that enables no driver.

## Affected targets

- `AIRBOTG4AIO`
- `KAKUTEG4AIO`
- `MADFLIGHT_FC3`
- `LUXHDAIO-G4`

## Why removal (behaviour-preserving)

- The define matches no firmware guard, so it does nothing today — removing it changes no compiled behaviour.
- None of these targets wire an SPI baro bus, so `USE_BARO_SPI_LPS` (SPI-only) was not the intent.
- `MADFLIGHT_FC3` already declares `USE_BARO_LPS22DF`, so its LPS coverage is unaffected.

If any of these boards physically carries an LPS22 pressure sensor on I2C, the correct fix is to add `USE_BARO_LPS22DF` (not the removed `USE_BARO_LPS`); this can be done by a maintainer with knowledge of the fitted hardware.

## How it was found

Surfaced while auditing barometer defines for the betaflight.com manufacturer design guidelines against firmware master and the config repo.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated supported barometer options on several boards, removing one sensor variant and adding broader compatibility on one configuration.
  * Some board setups now rely on alternative barometer support, which may improve hardware matching and reduce configuration issues.


osirisinferi (approved):

osirisinferi:
@coderabbitai review

ot0tot (approved):

PR #1142, opened 2026-07-06
JHEMCU add support for PY25Q128HA

haslinghuis:
## Summary by CodeRabbit

* **New Features**
  * Added support for a new flash memory option across multiple board configurations.
  * Expanded build-time hardware selection so more boards can use the updated flash variant.


osirisinferi (approved):

osirisinferi:
I'm fine with explicitely adding these defines *everywhere*, but note that the only thing `USE_FLASH_PY25Q128HA` actually does is pulling in `USE_FLASH_M25P16` in `common_post.h`. So for boards that already have `USE_FLASH_M25P16` defined, ultimately *functionally* nothing changes, this flash IC should already work for them. Doesn't hurt to add `USE_FLASH_PY25Q128HA` either though.

ot0tot (approved):

PR #1147, opened 2026-07-10
Add BARO (DPS310) and change ACC/GYRO to ICM20602 for ELINF405/722

neoomni:
### Description
This pull request updates the hardware configurations for both **ELINF405** and **ELINF722** targets to reflect the latest hardware specifications. 

It adds support for the onboard barometer and updates the default IMU configuration.

### Changes
- **IMU Update**: Changed `ACC` and `GYRO` drivers from `MPU6500`/`MPU6000` to `ICM20602` via SPI.
- **Barometer Addition**: Enabled `USE_BARO` and added `USE_BARO_DPS310` using `I2CDEV_1`.
- **Cleanup**: Removed unused `//TODO` OSD configurations and serial definitions to clean up the target header files.

### Target(s) Affected
- `configs/ELINF405/config.h`
- `configs/ELINF722/config.h`

### Testing
- Confirmed compilation compiles successfully for both targets.
- Verified sensor definitions align with the hardware layout.

## Summary by CodeRabbit

* **New Features**
  * Added support for ICM20602 accelerometer and gyroscope sensors on ELINF405 and ELINF722 boards.
  * Enabled DPS310 barometer support using the board’s I2C connection.
  * Enabled MAX7456 display support on ELINF405.

* **Chores**
  * Removed obsolete commented configuration placeholders from both board configurations.


ot0tot:
These are old targets. Were any FCs shipped with the MPU gyro? If so, the existing defines should not be removed.

neoomni:
https://betaflight.com/docs/wiki/boards/archive/ELINF405
https://betaflight.com/docs/wiki/boards/archive/ELINF722

According to the official documentation in the links above, these targets are specified to use the ICM-20602. As far as I know, these boards have used this IMU from the beginning, and the board I own also has the ICM-20602 IMU.

Since the MPU6500 and MPU6000 definitions do not seem to be used for these targets, I removed them to clean up the configuration.

ot0tot:
> https://betaflight.com/docs/wiki/boards/archive/ELINF405 https://betaflight.com/docs/wiki/boards/archive/ELINF722
> 
> According to the official documentation in the links above, these targets are specified to use the ICM-20602. As far as I know, these boards have used this IMU from the beginning, and the board I own also has the ICM-20602 IMU.
> 
> Since the MPU6500 and MPU6000 definitions do not seem to be used for these targets, I removed them to clean up the configuration.

Aren't you the manufacturer? If these boards have always shipped with the ICM-20602, how did they work without that define?

neoomni:
> Aren't you the manufacturer? If these boards have always shipped with the ICM-20602, how did they work without that define?

The manufacturer asked me to look into and fix an issue where the barometer was not being recognized, so I have been testing these changes for them.

When I tested the board with the current release version (2025.12.5), the IMU was successfully recognized as the ICM-20602 and functioned perfectly. To be honest, I am also not entirely sure how it managed to work without these definitions until now—it is quite strange.

Since my main goal for this PR is simply to get the BARO sensor working properly, do you think it would be better to leave the existing MPU6500 and MPU6000 configurations untouched just in case?

nerdCopter:
> [!NOTE]
> betaflight/config moved from `configs/<BOARD>/` to `configs/<MANUFACTURER_ID>/<BOARD>/` (grouped by manufacturer). This PR's `config.h`/`config.mk`/`config.c` are still at the old flat path and need to move to `configs/<MANUFACTURER_ID>/<BOARD_NAME>/` before this can merge. Rebase onto master, move the files, and confirm `MANUFACTURER_ID` is declared and matches the parent directory (`BOARD_NAME` must still match the leaf directory name). See the repo README for the current layout and required defines.

also, manufacturer need to confirm if old IMU was released. if valid, then cannot remove old IMU

PR #1151, opened 2026-07-14
Remove unneeded defines

ot0tot:
BMI270 does not support CLKIN

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the AXISFLYINGH743PRO_BMI270 board configuration to disable unsupported external gyro clock input.
  * Adjusted timer assignments to prevent conflicts with the disabled gyro clock pins.


haslinghuis (approved):

osirisinferi (approved):
You're right 🙂
Strictly speaking it's not necessary to remove them.. More than enough flash on the H7 and the CLKIN feature isn't enabled on IMUs other than ICM-42688-P, but I agree it's cleaner to not having them in this target.

PR #1159, opened 2026-07-18
Fix AXISFLYINGH7AIO UART4/UART5

osirisinferi:
In the schematic, the 'RX5' net, connected to PC11, is used for ESC telemetry from the AIO ESCs. The 'TX5' net is not used anywhere. However, PC11 only has USART3_RX or UART4_RX as alternate function, not UART5_RX. So let's change it to UART4 to fix this.
    
(Note that ESC telemetry is currently not working on this FC, as AM32 has removed serial telemetry output from the PA14 GPIO on non-F421 MCUs due to the dual role as GPIO and SWCLK this pin has. This might change in the future, but with current stock AM32 FW, ESC telemetry does not work. (For more info see https://github.com/am32-firmware/AM32/pull/373.) I have confirmed ESC telemetry working using UART4 on PC11 with a custom AM32 build implementing the aforementioned PR.)

### DIFF
--- a/configs/AXISFLYINGH7AIO/config.h
+++ b/configs/AXISFLYINGH7AIO/config.h
@@ -48,7 +48,6 @@
 #define UART1_TX_PIN         PA9
 #define UART2_TX_PIN         PD5
 #define UART3_TX_PIN         PD8
-#define UART5_TX_PIN         PC12
 #define UART6_TX_PIN         PC6
 #define UART7_TX_PIN         PE8
 #define UART8_TX_PIN         PE1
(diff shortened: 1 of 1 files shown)


haslinghuis (approved):

ot0tot (approved):
Regarding the AM32 telemetry issue, doesn't it support DShot EDT?

osirisinferi:
> Regarding the AM32 telemetry issue, doesn't it support DShot EDT?

Sure, but that's besides this issue. Everyone should use bidirectional DShot (with or without EDT) anyway. But the AIO ESCs are connected to `PC11` anyway, so why not make it actually work :slightly_smiling_face:

ot0tot:
> > Regarding the AM32 telemetry issue, doesn't it support DShot EDT?
> 
> Sure, but that's besides this issue. Everyone should use bidirectional DShot (with or without EDT) anyway. But the AIO ESCs are connected to `PC11` anyway, so why not make it actually work 🙂

Missed the **AIO** part 😅

PR #1178, opened 2026-08-07
Remove Pin IO user configurable USE flags

osirisinferi:
The `USE_PINIO` is a user configurable setting in the flasher and `USE_PINIOBOX` is enabled automatically if `USE_PINIO` is defined. Having `USE_PINIO` in the target `config.h` leads to build errors if the user enables "Pin IO" in the flasher, thus they need to be removed.

Technically we could also gate the `USE_PINIO` flag behind a `ifndef` to force it for specific boards, but we've never done that before. Adding `USE_PINIOBOX` is superfluous, so can be removed safely.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated flight-controller configurations to remove obsolete pin-control options.
  * Preserved video-transmitter enable-pin functionality for supported hardware.

### DIFF
--- a/configs/SPRO/SPRACINGH7EF/config.h
+++ b/configs/SPRO/SPRACINGH7EF/config.h
@@ -184,7 +184,9 @@
 #define ADC_EXTERNAL1_PIN    CURRENT_METER_2_ADC_PIN
 
 #define VTX_ENABLE_PIN                PB11
+#define USE_PINIO
 #define PINIO1_PIN                    VTX_ENABLE_PIN
+#define USE_PINIOBOX
 
 #define USE_ACC
 #define USE_ACC_SPI_ICM42605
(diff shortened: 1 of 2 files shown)


osirisinferi:
@coderabbitai Note that you should have flagged the usage of user configurable USE flags in the PRs for those targets when reviewing to begin with, but apparently they slipped through. Do you have a list of USE flags that are user configurable and thus you should flag as not to be added to target `config.h` files? Otherwise you would not be able to flag such errors.

osirisinferi:
@coderabbitai Very good. Now, in addition to the stuff you already mentioned yourself, note that for *some* of these defines, it's perfectly fine to `ifndef` gate the `#define`, such as for `USE_WING` for wing specific targets. Additionally, the same is true for e.g. `USE_MAG` if the FC has an onboard magnetometer. Thus you should be carefull regarding how strongly you would react to the presence of these user configurable flags, but if they appear in a target, they a) should always be gated and b) should be flagged for specific human review.

osirisinferi:
> If a target `config.h` defines one of these macros, it must use an `#ifndef` gate. This permits a Flasher build option to override the target default.

@coderabbitai The other way around: it permits the target override and force the lack of a Flasher build option.

haslinghuis (approved):

nerdCopter (approved):

PR #1181, opened 2026-08-08
DAKEFPV removed the LSM6DSV16X and added the BMI270

engineer-dakefpv:
Removed the LSM6DSV16X and added the BMI270

## Summary by CodeRabbit

- **New Features**
  - Added BMI270 accelerometer and gyroscope support across nine DAKE flight controller configurations.
  - Updated affected targets to use BMI270 sensor hardware instead of LSM6DSV16X.
  - Enabled BMI270 alignment compatibility with ICM-based configurations, improving sensor orientation support across the updated targets.


haslinghuis (changes requested):
BMI270 won't work without changing alignment as it does not use a dedicated footprint. Also it changes a IMU with another non-recommended IMU.

osirisinferi:
> BMI270 won't work without changing alignment as it does not use a dedicated footprint.

Technically, we could use [#15484](https://github.com/betaflight/betaflight/pull/15484) for this issue. Although it's only available in 2026.6.1 and not before.

> Also it changes a IMU with another non-recommended IMU.

Agreed, `USE_ACCGYRO_LSM6DSK320X` is already defined, recommend that IMU to be used over the BMI270.

engineer-dakefpv:
@haslinghuis  Currently, we default to using the LSM6DSK320X, but recently some customers have requested to use the BMI270 instead of the K320 gyro when purchasing flight controllers. For these flight controllers, at this stage, we can compile firmware with the BMI270 orientation adjusted. At the same time, we hope that customers can manually flash the firmware through the ground station after receiving the product (although the BMI270 orientation isn’t correct, it can still work after manual adjustment. Also, it seems this issue will be resolved after #15484).

haslinghuis:
@engineer-dakefpv - you need to add `#define ENABLE_BMI270_ALIGN_AS_ICM 1` after where the BMI is defined if it is sharing the footprint with the TDK / ST. Normally BMI270 would have its dedicated footprint. Updating manufacturer documentation.

haslinghuis (approved):

osirisinferi (approved):
Note that `ENABLE_BMI270_ALIGN_AS_ICM` is introduced in 2026.6.1 and is not present in earlier versions. So users flashing an earlier BF version on BMI270 boards would manually need to re-orientate their board alignment.

PR #1210, opened 2026-09-16
fix(ADC1/TIM8-CH3 DMA collision): STELLARF7, STELLARF7V2, STELLARF7_DUAL_HD, ELLE0

### LINKED ISSUE #1209: STELLARF7 / STELLARF7V2: ADC1 shares DMA2_Stream4 with TIM8 CH3, causes silent motor-DMA failure under dshot_burst=OFF
AI Generated issue-ticket

## Boards affected

STBE/STELLARF7, STBE/STELLARF7V2 (both STM32F722, identical wiring except motor pin order).

## Root cause

Both boards' config.h set:
- `ADC1_DMA_OPT 1`
- `TIMER_PIN_MAP` option 1 for the TIM8 CH3 motor pin (PC8 on STELLARF7, PC8 on
  STELLARF7V2)

Decoded against `src/platform/STM32/dma_reqmap_mcu.c`'s shared F4/F7 tables:

    { DMA_PERIPH_ADC, ADCDEV_1, { DMA(2, 0, 0), DMA(2, 4, 0) } },   // opt1 = Stream4
    { TIM8, TC(CH3),           { DMA(2, 2, 0), DMA(2, 4, 7) } },   // opt1 = Stream4

Both peripherals resolve to DMA2_Stream4 on the current config.

## Failure mode

`adc_stm32f7xx.c`'s `adcInit()` claims its configured stream via a single-shot
`dmaAllocate()` call with no retry/fallback if the claim fails. Firmware init order
runs ADC init before motor/PWM init, so ADC1 wins Stream4 first; the motor driver's
later `dmaAllocate()` call for TIM8 CH3 then fails, and that motor output gets no
DShot DMA configured -- no error is surfaced, the motor is simply inert.

## Reachability

Both boards default to `DEFAULT_DSHOT_BURST DSHOT_DMAR_ON`, which uses a separate,
non-colliding DMA claim path (the timer's shared "update event" request, not the
per-channel one). The collision is latent under default configuration and only
reachable if a pilot explicitly sets `dshot_burst = OFF`.

## Recommended fix

TIM8 CH3 has no alternative -- its only other option (Stream2) is permanently
shared with TIM8 CH1 on this table, so it cannot move. ~~ADC1's only other
documented option is Stream0. Checked both boards' full config.h for any other
DMA2_Stream0 consumer: single ADC instance (no ADC2/ADC3 use), no `USE_SDCARD`,
no `LED_STRIP_PIN` defined on either board. SPI1/SPI2/SPI3 (gyro/OSD/flash) go
through the generic SPI DMA path, which retries a second option on conflict, so
they are not a fixed claimant on any specific stream.~~

~~`-#define ADC1_DMA_OPT 1`~~
~~`+#define ADC1_DMA_OPT 0`~~

~~Same change on both STELLARF7 and STELLARF7V2.~~

**Correction (see comment below from @osirisinferi):** the struck-through reasoning
is wrong. SPI1_SDI's fallback DMA option is DMA2_Stream2, already permanently held by
TIM8 CH1, so a retry lands on Stream0 too -- colliding with the proposed ADC1 move.
ADC2's two options collide with TIM8 CH1/CH2 the same way. Only ADC3's second option
(DMA2_Stream1) is free.

Corrected fix: move the ADC off ADC1 onto ADC3, using ADC3's second DMA option. Both
PC0 and PC1 support ADC3, so no pin change is forced.

    -#define ADC_INSTANCE ADC1
    -#define ADC1_DMA_OPT 1
    +#define ADC_INSTANCE ADC3
    +#define ADC3_DMA_OPT 1

Same change on both STELLARF7 and STELLARF7V2.

nerdCopter:
AI Generated pull-request

## Summary

Fixes #1209: `ADC1` (`ADC1_DMA_OPT 1` -> DMA2_Stream4) and TIM8 CH3's motor DMA
(`TIMER_PIN_MAP` option 1 on the CH3 pin, also Stream4) collide on `STELLARF7`,
`STELLARF7V2`, `STELLARF7_DUAL_HD`, and `ELLE0` -- confirmed via a purpose-built resolver
tool that parses each board's real DMA/timer tables instead of pattern-matching, not just
the four boards a community candidate search happened to flag. `adcInit()` claims the
stream first at boot, so the motor's later DMA claim silently fails -- no DShot output on
that motor, no error surfaced.

Reachability differs per board: the first three are latent unless a pilot sets
`dshot_burst = OFF` (default `DSHOT_DMAR_ON` uses a different, non-colliding claim path).
`ELLE0` has no `DEFAULT_DSHOT_BURST` override at all, and this codebase's fleet-wide default
is `DSHOT_DMAR_OFF` -- the bug is live on `ELLE0` out of the box, not latent.

`STELLARF7_DUAL_HD` and `ELLE0` weren't in the original report; found and audited in
follow-up passes after this PR was opened, each confirmed via the boards' own DMA usage, not
assumed from the collision pattern alone.

## Fix

The right fix differs per board depending on what else is on DMA2 -- not the same patch
copy-pasted four times:

- **STELLARF7 / STELLARF7V2 / STELLARF7_DUAL_HD**: move the ADC off ADC1 onto ADC3
  (`ADC_INSTANCE ADC3`, `ADC3_DMA_OPT 1`), not onto ADC1's other option (Stream0) as first
  proposed on the issue. Stream0 is not actually free on these boards: SPI1's MISO DMA claim
  (`spiInitBusDMA()`) tries Stream0 first and Stream2 second, and Stream2 is permanently
  pinned to TIM8 CH1 -- moving the ADC there trades the ADC/motor collision for an
  ADC/gyro-SPI one. ADC3's second option (Stream1) is unclaimed by anything else on any of
  the three boards' DMA2 usage. PC0/PC1 both support ADC3 on this MCU, so no pin change is
  needed.
- **ELLE0**: this board has no SPI1 at all (single-gyro SPI2 only), so ADC1's other option
  (Stream0) genuinely is free here -- `ADC1_DMA_OPT 1` -> `0` is sufficient, no need to move
  ADC instances.

## Per-board verification (not pattern-matched)

Built a resolver that parses each board's config against the real
`src/platform/STM32/timer_stm32f{4,7}xx.c` (pin -> physical timer/channel) and
`dma_reqmap_mcu.c` (channel -> DMA stream) tables at run time, rather than reasoning about
scattered tables by hand for each board. Used it to:
- Confirm each of the four fixes above resolves cleanly (no new collision introduced) before
  writing any of them.
- Regenerate the community's F405/F722+`ADC1_DMA_OPT`+motors-PC6-PC9-in-order candidate
  search live against this repo: 22 boards, not the ~40 informally reported. Of those 22,
  only `STELLARF7` and `ELLE0` are real; the other 20 have `ADC1_DMA_OPT` present but set to
  a non-colliding value, which that search's presence-only check missed. The search's fixed
  motor-pin-order assumption also structurally cannot find `STELLARF7V2` or
  `STELLARF7_DUAL_HD` (both wired in reverse order), which is exactly why this PR checks
  those independently rather than trusting that search's coverage.

## Verification

Built all four configs against a fresh `betaflight/betaflight` master worktree with
`make EXTRA_FLAGS=-Werror CONFIG=<board> BETAFLIGHT_CONFIG=<this branch>` -- the same
invocation this repo's own CI (`pr.yml`) uses -- all clean, no warnings.

`STELLARF7`/`STELLARF7V2` also cross-checked against the equivalent fix already applied and
build-verified on the EmuFlight fork of this codebase (`emuflight/EmuFlight#1454`, same root
cause, different DMA-selection mechanism on that fork). EmuFlight has no equivalent of
`STELLARF7_DUAL_HD`. EmuFlight's `ELLE0` has the identical, live bug -- not yet fixed there,
tracked separately.


nerdCopter:
request @cvetaevvitaliy review 
(i do not own this hardware)

ot0tot:
This is neither an issue or a valid solution. Please do not act as a meatproxy for AI 😅

